PR #115, opened 2013-11-01
configurable wrapping of single links/resources

Laures:
Hal defines the json-array around single links or embedded resources as optional. currently the hal implementation will never wrap single entities. now this can be activated.

Details:
- wrapping is off by default
- can be configured on the handler instantiator


gregturn:
Superceded by #288.

PR #197, opened 2014-06-11
Added ResourceMixin and HalResourceSerializer

### LINKED ISSUE #191: Resource of JsonNode not deserializable from Hal
Given a JsonNode which represents a Person.
When I serialize it as a `Resource<JsonNode>`, the JsonNode is not unwrapped, but it is attached to a field "content". If I were using a `Resource<Person>`, the person attributes would be unwrapped directly on the Json object, without a field "content".
If I then try to deserialize the Json object with "content" to a `Resource<Person>`, the fields of the resulting person are null.

I have taken it already, pull request follows.

dschulten:
Fixes #191. A side-effect of this change is that the order of attributes on embedded beans is now links-first, but that seems acceptable since the example HAL documents use the same ordering.

### DIFF
--- a/src/main/java/org/springframework/hateoas/Resource.java
+++ b/src/main/java/org/springframework/hateoas/Resource.java
@@ -38,7 +38,7 @@ public class Resource<T> extends ResourceSupport {
 	/**
 	 * Creates an empty {@link Resource}.
 	 */
-	Resource() {
+	protected Resource() {
 		this.content = null;
 	}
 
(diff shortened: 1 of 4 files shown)


pivotal-issuemaster:
@dschulten Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=197)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=197) to manually synchronize the status of this Pull Request.

gregturn:
Not seeing any expressed need on this for three years, I think this is unlikely.

Issue #209, opened 2014-07-03
ControllerLinkBuilder fails on Map<> RequestParams


rossturner:
I have a Controller that I would like to accept any request parameters on. According to [@RequestParam](http://docs.spring.io/spring/docs/4.0.4.RELEASE/javadoc-api/org/springframework/web/bind/annotation/RequestParam.html) javadoc:

> If the method parameter is Map<String, String> or MultiValueMap<String, String> and a parameter name is not specified, then the map parameter is populated with all request parameter names and values.

So I should be able to accept either of these types in a Controller in this way to accept any query parameters. However, `ControllerLinkBuilderFactory.linkTo` either deals with `instanceof Collection` arguments or else attempts a `.toString()` on the parameter which is failing.

I'd like to extend this to also check for `instanceof Map<String, String>` and `instanceof MultiValueMap<String, String>` and add a set of queryparams to the builder. If this seems like a reasonable change I'm happy to submit a pull request.

PR #210, opened 2014-07-03
#209 - Added support for Map<String, String> and MultiValueMap<String, S...

rossturner:
See [https://github.com/spring-projects/spring-hateoas/issues/209].

I've added unit tests to cover `@RequestParam` Controller arguments with no name that are Map<String, String> or MultiValueMap<String, String>. The only thing I'm not really happy about in this commit is the untyped cast from `Map` to `Map<String, String>` (and similarly for the other type) but as only these generified types are accepted I don't believe it's necessary to check for other types or add an exception handler around the cast.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
 
 	<groupId>org.springframework.hateoas</groupId>
 	<artifactId>spring-hateoas</artifactId>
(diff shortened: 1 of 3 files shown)


olivergierke:
Applied with polishing.

PR #236, opened 2014-08-24
Add means to render empty embedded collections

elnur:
An API should be consistent. If when we have elements in a collection, we get a response like

``` json
{
    "_embedded": {
        "users": [
            {
                "username": "bob"
            },
            {
                "username": "joe"
            }
        ]
    }
}
```

and then when there are no users found, we get a response like

``` json
{
}
```

it is not consistent and forces our clients to handle the special case of not having `_embedded.users` in a response.

Instead, what our clients reasonably expect is a response like

``` json
{
    "_embedded": {
        "users": [
        ]
    }
}
```

### DIFF
--- a/src/main/java/org/springframework/hateoas/hal/ResourcesMixin.java
+++ b/src/main/java/org/springframework/hateoas/hal/ResourcesMixin.java
@@ -30,7 +30,7 @@ public abstract class ResourcesMixin<T> extends Resources<T> {
 	@Override
 	@XmlElement(name = "embedded")
 	@JsonProperty("_embedded")
-	@JsonSerialize(include = JsonSerialize.Inclusion.NON_EMPTY, using = Jackson2HalModule.HalResourcesSerializer.class)
+	@JsonSerialize(using = Jackson2HalModule.HalResourcesSerializer.class)
 	@JsonDeserialize(using = Jackson2HalModule.HalResourcesDeserializer.class)
 	public abstract Collection<T> getContent();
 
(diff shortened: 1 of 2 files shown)


olivergierke:
Looks my fix for #286 already puts all the means in place you need. I've pushed a test case in 629028a. Would you mind taking the snapshots for a spin?

elnur:
That looks good, @olivergierke. I actually had no idea about embedded wrappers. And I now want to figure out how to use them. Would be really nice to have this in the docs.

elnur:
So, @olivergierke, can I use embedded wrappers to add embedded elements to a normal `Resource`? Seems like there's either a `Resource` without `_embedded` or `Resources` with everything in `_embedded`. I need both.

elnur:
I would really love to at least find a way to work around this problem.

eonardol:
I have solved in this way:

    EmbeddedWrappers wrappers = new EmbeddedWrappers(false);
    EmbeddedWrapper wrapper = wrappers.emptyCollectionOf(MyBean.class);

For PagedResources result

    List<EmbeddedWrapper> embedded = Collections.singletonList(wrapper);
    PagedResources pagedResources = new PagedResources(embedded, metadata, links);

For Resources result

    Resources<Object> resources = new Resources<>(Arrays.asList(wrapper));

gregturn:
Closing this issue since the original problem was solved via #286. If there is a new need, please open a new issue.

Issue #229, opened 2014-08-22
HalEmbeddedBuilder needs to support curie for the embedded resources


jstano:
The change i simple, just need to pass the CurieProvider into the class. I've creating a Pull Request with the changes.

PR #239, opened 2014-08-26
#229: added curie support to the HalEmbeddedBuilder class

jstano:


olivergierke:
That already has been applied with 3ca6fa5504c14cbe04339603efb3a991d9f9f742.

Issue #235, opened 2014-08-24
Add support to the Link class for the RFC 5988 attributes


jstano:
Add direct support for the following attributes:
   anchor, hreflang, media, title, type using the standard with builder style method

In order to support link extensions, we could do something like
   withExtension(String paramName, String paramValue)

I will submit a pull request with changes to support this enhancement.

olivergierke:
Fixed with #567.

PR #240, opened 2014-08-26
#235: added support for RFC 5988 Target Attributes to the Link class

jstano:


olivergierke:
Again, I'd love to hear about the rationale for this change. "I'd like to see the attributes" is not gonna cut it. Be sure to understand that we're basically a mitigating library that needs to be careful not to add attributes that the media types we support don't support in turn. Especially if we run into scenarios with conflicting or ambiguous attribute semantics.

Going forward, please make sure you add a few more explanations with your pull requests as simply throwing some code over the fence is not gonna work out anyway.

jstano:
I would expect that any hateoas media type that included links would be required to support all RFC-5988 target attributes, since this is the official standard for links. RFC-5988 also allows links to have custom attributes added.

Yes, I'll be sure to put for details in from now on.

Vulcannis:
This is a prerequisite for handling secondary key lookups in HAL. Even without that feature, if the attributes aren't parsed at all they can't be inspected by client code manually. This really cripples link collections.

gregturn:
According to the [HAL spec](https://tools.ietf.org/html/draft-kelly-json-hal-08#section-5), the only required attribute is **href**. All others are optional. 

We have presently implemented **templated** because it's critically needed for any controller with arguments. 

Something more is needed for any of the other attributes to be added. Otherwise, they can easily appear as superfluous bits.

pivotal-issuemaster:
@jstano Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=240)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=240) to manually synchronize the status of this Pull Request.

olivergierke:
Superseded by #567.

Issue #242, opened 2014-08-26
Add a resources property to the Resource class


jstano:
It's not clear how to get the _embedded stuff in the Hal renderer working without having a new property in the Resource class or having a class derived from Resource with a resources property.

If you can provide an example of how to accomplish this without any changes to the project, that might be helpful.

Here's what I'm proposing:
   new Resource(someDTO)
      .withResources(someCollectionOfResources)
      .withResources(otherCollectionOfResources)

If no resources are added to the Resource, then nothing will be output by the renderers.

I'll create a Pull Request so we look over the code.

gregturn:
Resolved via https://github.com/spring-projects/spring-hateoas/issues/864.

PR #243, opened 2014-08-26
#242: added a resources property to the Resource class in order to more ...

jstano:
...easily support a resource having embedded or nested resources

### DIFF
--- a/src/main/java/org/springframework/hateoas/Resource.java
+++ b/src/main/java/org/springframework/hateoas/Resource.java
@@ -15,8 +15,11 @@
  */
 package org.springframework.hateoas;
 
+import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


pivotal-issuemaster:
@jstano Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=243) to manually synchronize the status of this Pull Request.

gregturn:
I don't like this design given how it breaks the OO structure of the `ResourceSupport` type hierarchy.

Given we support multiple mediatypes, not all of which support the concept of embeddables, we need some sort of API that allows this to be done but without forcing the knowledge of HAL into these types.

PR #245, opened 2014-08-26
244: added methods to the EntityLinks interface and associated implement...

jstano:
...ation classes to make using the API a touch less verbose


olivergierke:
I don't think we're going to apply the proposed changes. Being aware of details of the controller managing an instance basically subverts the purpose of the `EntityLinks` interface in the first places. It solely exists for you to be able to create links with nothing known but the entity type.

PR #255, opened 2014-09-24
Upgrade to spring 4.1.0.RELEASE

dilipkrish:
Not really sure if there is a reason this project is lagging behind the release train. Seemed simple enough to upgrade.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -58,7 +58,7 @@
 
 	<properties>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-		<spring.version>3.2.9.RELEASE</spring.version>
+		<spring.version>4.1.0.RELEASE</spring.version>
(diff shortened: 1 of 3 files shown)


olivergierke:
Superseded by #311, upgraded to Spring 4.0.9.RELEASE.

dilipkrish:
Why 4.0.9 vs 4.1.x?

olivergierke:
The policy (mostly originating from Spring Data modules) is to be on the last but one major version. A reasonable balance to not cut off users that might not be able to upgrade to the latest and greatest (due to whatever reason) but still stay relatively up to date.

Note, that the version is only a lower baseline for us to build against. We make sure, Spring HATEOAS is  working fine on more recent versions of Spring Framework.

dilipkrish:
:+1: Thanks!

Issue #201, opened 2014-06-14
Make RestTemplate of Traverson configurable


dschulten:
The Resttemplate of Traverson and the underlying httpclient is encapsulated so that it is not possible to customize it, e.g. for SSL, various authentication schemes, caching etc.
Allow this, maybe by adding a ctor that takes a RestTemplate which is picked up and enhanced by adding the necessary configuration to it. 
Please give some feedback if that approach is ok for you so I can implement a fix, or if you would do that differently.

PR #258, opened 2014-10-08
Expose Traverson default message converters in a convenience method

dsyer:
The existing encapsulation of RestTemplate (via #201, #203) makes it the 
user's responibility to set up the correct message converters (which as argued
in #203, is necessary), but it doesn't give any help knowing what the
converters should be.

This change exposes the default message converters via a static convenience
method, e.g.

```
RestTemplate restTemplate = new RestTemplate();
restTemplate.setInterceptors(
    Arrays.<ClientHttpRequestInterceptor>asList(interceptor));
restTemplate.setMessageConverters(
    Traverson.getDefaultMessageConverters(MediaTypes.HAL_JSON));
```

### DIFF
--- a/src/main/java/org/springframework/hateoas/client/Traverson.java
+++ b/src/main/java/org/springframework/hateoas/client/Traverson.java
@@ -70,9 +70,11 @@ public class Traverson {
 	}
 
 	private final URI baseUri;
+
 	private final List<MediaType> mediaTypes;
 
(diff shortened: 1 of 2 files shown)


olivergierke:
Merged with minor changes.

gregturn:
Resolved via 912337669a4f3b34695f705ea0996fbaec66c856

PR #272, opened 2014-12-14
Fix reference to non-existing interface

amr:
It mentions LinkProvider where it meant LinkDiscoverer


olivergierke:
Applied, thanks!

PR #273, opened 2014-12-23
Fix Assert, forgot to pass the parameter that was being checked.

fleaflicker:


olivergierke:
I've added a unit test and merged your fix, thanks!

Issue #288, opened 2015-02-06
Allow links on a particular rel to be displayed as an array even if there is only one link


vivin:
Currently, `Jackson2HalModule.HalLinkListSerializer` will serialize a `rel` with only one link, into an object. If the `rel` has more than one link, it will serialize it into an array. 

I know that `rel`s have nothing to do with cardinality. However, for consistency's sake I would like to force certain a `rel` to always represent its links as being in an array. I ran into this issue in a case where I return a "collection" representation that contains links to individual resources. If I only have one element in the collection, the link is serialized as:

``` json
"ex:persons/person-resource": {
    "href": "..."
}
```

However, if I have multiple items, I will see:

``` json
"ex:persons/person-resource": [
    { "href": "..." },
    { "href": "..." }
]
```

From the perspective of an API user, this is remarkably inconsistent and confusing. I would rather that the representation not change based on the cardinality of elements. It is fine if the representation is always going to contain a link to a single element under some `rel`. In that case, it is alright that the link is serialized into a object (IMHO, it's a bit of a weakness because you don't see the same issue in HAL+XML where you can put a `<link>` element inside a `<links>` whether there is one or more `<link>`, but that's another issue). However in the case of a representation that is returning a collection of elements, I think it would be valuable from a consistency and usability perspective to always return the link as an array.

It would be nice if this was configurable when building the link somehow. A way, perhaps, to let the serializer know that the links should be serialized as an array instead of an object, regardless of cardinality.

Also, the HAL specification says:

> **Note**: If you're unsure whether the link should be singular, assume it will be multiple. If you pick singular and find you need to change it, you will need to create a new link relation or face breaking existing clients.

Currently, there is no way to force a link to be multiple in Spring HATEOAS. I hope you guys will consider this feature request. I can also try and solve this problem myself and make a pull request.

If this is not feasible, or if I am misinterpreting the HAL specification, please let me know!

schvanhugten:
I think the problem is in Jackson2HalModule:OptionalListJackson2Serializer:355 to 366 (version 0.16).

```java
if (list.isEmpty()) {
    return;
}

if (list.size() == 1) {
	serializeContents(list.iterator(), jgen, provider);
	return;
}

jgen.writeStartArray();
serializeContents(list.iterator(), jgen, provider);
jgen.writeEndArray();
 ```

[...]

PR #289, opened 2015-02-07
issue #288: (Approach #1) Allow links on a particular rel to be displayed as an array even if there is only one link

vivin:
Allows links under a single rel to be represented as a multiple (i.e., serialized to an array) even if there is only one.

This is one, possible approach. I don't like this too much since it is adding a rather-specific concern to the `Link` object (i.e., one that only comes into play in the case of HAL+JSON). Also, is it the responsibility of the `Link` to convey the information that must be serialized into an array under the `rel` that it belongs?


vivin:
After reflecting on this further, I think this approach is less than desirable. `Link` is an abstract concept of a link and shouldn't have representation-specific concerns. I'm closing this PR.

Issue #247, opened 2014-09-02
Add HAL_JSON_VALUE in MediaTypes class


nowakp1:
As I mentioned in title it would be great to add HAL_JSON_VALUE in MediaTypes class - similar like MediaType Spring class (pure string value which could be used i.e. @RequestMapping).

PR #329, opened 2015-04-04
#247 - Added HAL_JSON_VALUE in MediaTypes class

pnowy:
It has been added the HAL_JSON_VALUE in MediaTypes class - similar like MediaType Spring class (pure string value which could be used i.e. @RequestMapping).


pnowy:
You are right. I've change it according with the convention from base spring media type.

Issue #331, opened 2015-04-13
Inconsistent behavior when using linkTo(methodOn()) on handler methods with defaultValue attribute set on @RequestParam


systemfreund:
According to the Spring documentation `RequestParam.required` is implicitly set to `false` when `RequestParam.defaultValue` differs from the default value. The HATEOAS component, however, fails with the exception below when I change `defaultValue` unless I also explicitly set `required=false`.

```
java.lang.IllegalArgumentException: Required controller parameter XXX of method [...] found but null value given!
	at org.springframework.hateoas.mvc.AnnotatedParametersParameterAccessor.verifyParameterValue(AnnotatedParametersParameterAccessor.java:95)
	at org.springframework.hateoas.mvc.ControllerLinkBuilderFactory$RequestParamParameterAccessor.verifyParameterValue(ControllerLinkBuilderFactory.java:236)
	at org.springframework.hateoas.mvc.AnnotatedParametersParameterAccessor.getBoundParameters(AnnotatedParametersParameterAccessor.java:70)
	at org.springframework.hateoas.mvc.ControllerLinkBuilderFactory.linkTo(ControllerLinkBuilderFactory.java:133)
	at org.springframework.hateoas.mvc.ControllerLinkBuilder.linkTo(ControllerLinkBuilder.java:136)
```

PR #332, opened 2015-04-13
Inconsistent behavior when using linkTo(methodOn()) on handler methods

systemfreund:
Fix for #331


olivergierke:
Thanks for that, that's merged! :+1:

PR #340, opened 2015-05-17
Feature/affordances

dschulten:
Initial version of Affordance and AffordanceBuilder for spring-hateoas. Still everything in the original packages and with the original copyright notice, as discussed. To be decided if a message converter using Affordance should be in spring-hateoas and if so, which one. Uber[1] and Html[2] are candidates, but maybe it would be good if we start a different converter to work with Affordance from scratch.

[1] http://tinyurl.com/nslmlha
[2] http://tinyurl.com/ojmqfkg


olivergierke:
Some initial observations:

- some of the stuff in the `action` package seems to be quite HTML-specific. Is that stuff we can still use generally speaking or does it have to be generalized even more?
- the annotation configuration style looks pretty nifty, we probably have to discuss the details in the call

dilipkrish:
@dschulten @olivergierke :+1: for pushing this forward. Let me know if I can help in any way. I'm interested in this from the angle of adding hypermedia support for swagger (unlike the url centric wall of operations approach that it provides of the box.

dschulten:
@dilipkrish it would be helpful if you start using the AffordanceBuilder for your intended purpose to see if the descriptors contained in the Affordance contain everything you need.

dschulten:
@olivergierke @aheusingfeld 
- the action package sounds HTML specific, but it isn't meant to be used by HTML only. I sticked to HTML wording since it is well-known, besides I certainly see HTML5 as a potential media type to render Affordances.
- are there existing annotations we could use instead of having our own?

gregturn:
@dschulten I'm interesting in working on this PR. Are you able to squash it all into commit and rebase against latest master?

gregturn:
More comments as I sift through the code:
* We should drop UBER_XML considering the whole project is shaped around JSON.
* `HypermediaTypes` should be merged into `MediaTypes` and adopt the same conventions (a `MediaType` implementation like `HAL_JSON` as well as a string version like `HAL_JSON_VALUE`).
* Siren needs some additions to catch up to Uber
   * We need a `Siren2JacksonHttpMessageConverter` that uses `SIREN_JSON`.
   * Uber has `UberMessageModel` but Siren does not.
   * UberUtils appears to have a lot of code that's not Uber-specific. Wonder if we can extract that and re-use it for SirenUtils.
   * UberJackson2HttpMessageConverter contains duplicate code with SirenJackson2HttpMessageConverter. Extracting that to a common class would be handy.
* Spring HATEOAS provides `TypeConstrainedMappingJackson2HttpMessageConverter` to convert HAL formatted messages.
```
MappingJackson2HttpMessageConverter halConverter = new TypeConstrainedMappingJackson2HttpMessageConverter(ResourceSupport.class);
halConverter.setSupportedMediaTypes(Arrays.asList(HAL_JSON));
halConverter.setObjectMapper(halObjectMapper);
```
This is a sub-sub-sub-class of `AbstractHttpMessageConverter<Object>`, the type used by Uber and Siren message converters. HAL is leveraging Jackson's mixin nature. Wondering exactly where we should unify this. One possibility is to adjust how Uber and Siren are encoded, using the same system. But I don't know if this is simpler or harder than the newly written way to encode. Inclined to think the new converters are inadequate, because they can only write, not read. The mixin approach ensures Jackson converters either way are built and registered in a module, so this may be best.

Using the Lombok Builder pattern, it may be easier to declare a format that way instead of building up a series of nodes. I would like to experiment with going that way, and hooking into the afforance API.

gregturn:
Superceded by #612 and resolved via https://github.com/spring-projects/spring-hateoas/commit/70448a854098343897fd99e3227b4a61db54b921.

Final solution isn't precisely the same thing, but we plan to iterate over it in the near future and expand to support other mediatypes. Feel free to check the HAL-FORMS section of code and their corresponding unit tests.

Also, see monitor [Spring HATEOAS Examples repo](https://github.com/spring-projects/spring-hateoas-examples) as we plan to add an Affordances example very soon.

pivotal-issuemaster:
@dschulten Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=340)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=340) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@dschulten Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=340)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=340) to manually synchronize the status of this Pull Request.

Issue #342, opened 2015-05-18
Augment error message when building links outside of an MVC request


gregturn:
linkTo(), methodOn(), etc. methods only work inside Spring MVC handlers because they rely upon gathering info from the current web request. Add a clearer error message when someone tries to use them elsewhere.

PR #344, opened 2015-05-18
#342 - Augment error message when building link without a web request

gregturn:


olivergierke:
That's applied with some polishing.

Issue #354, opened 2015-06-03
Automated deployment of snapshots


gregturn:
Every time a merge is made to master, do a snapshot release.

gregturn:
Resolved via https://github.com/spring-projects/spring-hateoas/issues/764

PR #357, opened 2015-06-03
#354 - Automated deployment of snapshots

gregturn:
When commits are merged to master, automatically deploy via Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,4 +12,6 @@ cache:
     - $HOME/.m2
 sudo: false
 install: true
-script: "mvn clean dependency:list test -P${PROFILE} -Dsort"
+script:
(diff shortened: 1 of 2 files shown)


olivergierke:
I played with that and it doesn't quite do what I think we wanted to achieve. The binary snapshots are already deployed by the Bamboo build. To deploy the reference documentation – and I think that's we wanted to add here – we need to be in the SpringSource VPN, which unfortunately also the custom `settings.xml` can't get us in.

gregturn:
https://github.com/spring-cloud/spring-cloud-spinnaker/blob/master/pom.xml#L463-L482 shows how to upload artifacts and tap into Trevor's cron job which publishes artifacts into the target area.

gregturn:
Superceded by #764.

PR #358, opened 2015-06-08
Unable to deserialize Id Simple Pojo correctly

mfalaize:
Here is a test case for my issue. It seems we cannot deserialize an Id simple pojo with PagedResources<Resource> type.

What do you think about it ?

See also the issue on [stackoverflow](http://stackoverflow.com/questions/30695182/null-id-property-when-deserialize-json-with-jackson-and-jackson2halmodule-of-spr).

### DIFF
--- a/src/test/java/org/springframework/hateoas/hal/IdSimplePojo.java
+++ b/src/test/java/org/springframework/hateoas/hal/IdSimplePojo.java
@@ -1,0 +1,14 @@
+package org.springframework.hateoas.hal;
+
+public class IdSimplePojo extends SimplePojo {
+
+    private Integer id;
+
+    public Integer getId() {
+        return id;
+    }
+
+    public void setId(Integer id) {
+        this.id = id;
+    }
+}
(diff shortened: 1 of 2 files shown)


mfalaize:
Actually it is because of the Resource class that contains a getId method annotated by @JsonIgnore. As the content property in the json is deserialized by an unwrapper *after* the id property was ignored, it cannot handle the id property.

I'll think about a patch soon.

mfalaize:
It seems that the id property of the ResourceSupport class has to be ignored only in serialization. We can easily fix the error with the last version of Jackson (at this time 2.6.0-rc2) with removing the @JsonIgnore on the getId() method and adding @JsonIgnoreProperties({ "id" }, allowSetters=true) to the ResourceSupport class. AllowSetters is a newly created property of this annotation in the 2.6.0 jackson version but when we launch unit tests with this version of jackson, many tests don't pass...

Is there another way to allow deserialization of the id property or are you ok to upgrade jackson version ?

olivergierke:
This has been brought up a couple of times: in REST the ID is the URI, thus calling `getId(…)` on a `ResourceSupport` will return the self link of the resource.

PR #372, opened 2015-07-30
LinkBuilderFactory linkTo parameter map support

Vorago:
JaxRsLinkBuilder has two linkTo methods:

```
linkTo(Class<?> service)
linkTo(Class<?> service, Object... parameters)
```

Yet there is no method for a map, which is extremely useful in case you are using UriInfo#getPathParameters() to extract parameter values.

Also, UriComponents has expand method that takes a map as an argument, so it seems logical to me to have another linkTo method:
```
linkTo(Clas<?> service, Map<String,?> parameters)
```

P.S.
Implementing this method only for JaxRsLinkBuilder wouldn't be right, so I also did it for LinkBuilderFactory and descendants


olivergierke:
That's merged, thanks!

PR #376, opened 2015-08-12
Compile against latest JAX-RS API

wilkinsona:
This PR updates Spring HATEOAS' optional dependency on the JAX-RS API to use the latest version (which uses a different artifact ID). This change will make things a little easier for the Spring IO
Platform which currently provides dependency management for Jersey 2.x (which uses JAX-RS 2.0), by allowing it to stop providing dependency management for the JAX-RS 1.0 API.


olivergierke:
That's merged, thanks!

PR #382, opened 2015-09-02
Fix to Traverson double encoding issue (337) plus test case

tommyb82:
@olivergierke @gregturn 

Hi guys, here's the PR with squashed commits, rebased onto current master (hope that's correct - first time doing some of this stuff for me..).  

It was a little more complicated in the end as my original change had broken another Traverson test case (`TraversonTest#returnsTemplatedLinkIfRequested`).  So there's now an additional private helper method in there to provide the option of traversing to a final, expanded `URI` (as required for the calls to `RESTOperations#exchange()`, or alternatively traverse to a templated String

Hope that makes sense, let me know what you think.


olivergierke:
Polished and merged. Thanks!

tommyb82:
Awesome, great to have contributed.

gregturn:
👍

PR #386, opened 2015-09-05
Add maven lifecycle mapping for springsource bundlor

tinne:
When using spring-hateoas with the eclipse ide and its integrated maven support m2e, the project leaves a warning that it does not know how to interact with the springsource OSGi bundlor plugin for the reasons explained [here] (https://www.eclipse.org/m2e/documentation/m2e-execution-not-covered.html).

This patch explains it to run the plugin on any build and removes the warning.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -556,6 +556,36 @@
 	</dependencies>
 
 	<build>
+		<pluginManagement>
+			<plugins>
+				<plugin>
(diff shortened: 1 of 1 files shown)


olivergierke:
This can be achieved by adding Bundlor to the list of ignored plugins in Eclipse. We're not gonna add IDE specific POM setup.

PR #390, opened 2015-09-20
Rename missing Tests to Test for maven support

gregturn:


olivergierke:
That's merged, thanks!

Issue #361, opened 2015-06-22
Add Support for Resolving Property Values in @RequestMapping(value)


kevinconaway:
The `@RequestMapping(value)` annotation supports property values for dynamically specifying the endpoint value, like `@RequestMapping(value="${some.request.mapping})`

I am using the [Springfox](http://springfox.github.io/) project to document my API and I would like to be able to add a Link to the [Swagger2 documentation](https://github.com/springfox/springfox/blob/fd503636177b518c07b55f8d323630f2ec6c431f/springfox-swagger2/src/main/java/springfox/documentation/swagger2/web/Swagger2Controller.java) in my API:

I'm adding a link to a resource via:
 `resource.add(linkTo(methodOn(Swagger2Controller.class).getDocumentation(null)).withRel("swagger.json"));` 

and the resuling link looks like:

 `http://localhost:8080/api/${springfox.documentation.swagger.v2.path:/v2/api-docs}`

jghiloni:
I'd like to throw my support behind this as well

joeybic:
Is anybody still working on this? I know it's been more than a year, but it'd be super helpful if this worked...

huehnerlady:
@gregturn Any update on this? It would be very important to us, will this be fixed soon? It is hanging for 3 years now 😞

odrotbohm:
Resolved by #1328

mirceade:
I don't get it, this doesn't really work in 1.4.0 with Kotlin.

PR #391, opened 2015-09-20
Issue #361: Add Support for Resolving Property Values

jghiloni:
Add Support for Resolving Property Values in @RequestMapping(value)

Creates extra methods in `AnnotationMappingDiscoverer`, `ControllerLinkBuilder`, and `ControllerLinkBuilderFactory` to take an `Environment` object that can be used resolve placeholders. Accepts null `Environment`s to allow for backward compatibility.

### DIFF
--- a/src/main/java/org/springframework/hateoas/core/AnnotationMappingDiscoverer.java
+++ b/src/main/java/org/springframework/hateoas/core/AnnotationMappingDiscoverer.java
@@ -20,12 +20,15 @@ import static org.springframework.core.annotation.AnnotationUtils.*;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.Method;
 
+import org.springframework.core.env.Environment;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 
 /**
  * {@link MappingDiscoverer} implementation that inspects mappings from a particular annotation.
  * 
  * @author Oliver Gierke
+ * @author Josh Ghiloni
  */
 public class AnnotationMappingDiscoverer implements MappingDiscoverer {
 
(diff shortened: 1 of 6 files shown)


jhoeller:
Could use `PropertyResolver` instead of `Environment`, or even a plain `StringValueResolver` that could be adapted from a `PropertyResolver` (see `PropertySourcesPlaceholderConfigurer`'s implementation).

jghiloni:
I like the idea, but I'm unclear as to where the `StringValueResolver` would get the values to use for substitution. Are you suggesting passing that in instead of an `Environment`, or somehow having the resolver as an instance variable?

jhoeller:
Indeed, a `StringValueResolver` would have to be adapted from a `PropertyResolver` along the lines of `PropertySourcesPlaceholderConfigurer`, line 170+, so it's not really convenient to use for that purpose. Declaring `PropertyResolver` itself in such a signature might be the better choice, since `Environment` implements `PropertyResolver` and can therefore be passed in directly.

jghiloni:
That's a good point. I'll make the necessary changes to reflect that (Passing `PropertyResolver` instead of `Environment`)

jghiloni:
Any news on this?

gregturn:
@jghiloni Ollie has been on PTO since SpringOne so things may be slow in coming out right now.

jghiloni:
Ah, makes sense! No real rush, just noticed it was still open.

olivergierke:
I don't like that now the `ControllerLinkBuilder` is now scattered with `PropertyResolver` parameters, something as it really breaks the abstraction (why would a client have to know about property resolution when calling those methods).

Maybe we can move the changes to the `ControllerLinkBuilderFactory`, which is an instance based alternative to the static `ControllerLinkBuilder`, which then can take an `Environment` as configuration setup to then use it during link resolution without polluting the parameter space?

gregturn:
Catching up on @olivergierke 's comments on this feature. I'll try to see how this now operational PR can be retooled to not percolate this change into so many APIs.

gregturn:
The related branch now shows a different solution that doesn't require altering any method signatures. It's also more composable.

PR #399, opened 2015-10-10
Added VndErrorException

jmnarloch:
Hi, 

Some time ago I had created this little thingy for handling the VndErrors in Feign calls over SpringCloud: https://github.com/jmnarloch/feign-vnderror-spring-cloud-starter. I though that I would be better if there would be a unified exception that could be handled by multiple clients: Feign, RestTemplate, HttpClient, OkHttp etc. So I would like to propose adding one.

### DIFF
--- a/src/main/java/org/springframework/hateoas/VndErrorException.java
+++ b/src/main/java/org/springframework/hateoas/VndErrorException.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2013-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.hateoas;
+
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpStatus;
+import org.springframework.web.client.HttpStatusCodeException;
+
+import java.nio.charset.Charset;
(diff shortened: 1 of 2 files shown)


jmnarloch:
cc @olivergierke

olivergierke:
What would that exception be used like? Would you mind adding test cases for the code you added? I guess we need to give the `VndErrors` implementation a bit of a rewrite anyway as it seems like the spec has moved significantly in the meantime.

jmnarloch:
So the whole idea is that this introduces a specialized version of HttpStatusCodeException that brings the payload information in a form of VndErrors, I am not expecting anything more that the Spring HATEOAS would just define it. This opens possibilities to try to unify the errors handling using different clients API like already mentioned: RestTemplate, AsyncRestTemplate and for instance Feign clients, but the integration would have be handled elsewhere. This would bring at least the common error representation.

aheusingfeld:
I like the idea but could you explain how you avoid that the libraries you mentioned (Feign, RestTemplate, HttpClient, OkHttp etc.) need to have a dependency to the artifact which contains VndErrorException i.e. speing-hateoas?

jmnarloch:
Probably we can not avoid that, I made the PR here were the VndErrors is already being defined, though I would consider whether it wouldn't make more sense to move the VndErrors to spring-web at some point.

jmnarloch:
@olivergierke Thanks, for your thoughts. I've added extra tests, if you want to I can take a look at updating the implementation and align the VndErrors with the current specs, though I would rather prepare separate PR for that.

Also I was wondering way the exception hasn't been introduced in first place, the spring-web has this hall hierarchy of RestClientException.

pivotal-issuemaster:
@jmnarloch Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=399) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@jmnarloch Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=399)!

Issue #143, opened 2014-01-20
How let ControllerLinkBuilder see original request URI?


andrewtswu:
Hi,

Our tomcat server can only see internal URIs due to the way we setup our servers. If a client, for example, wants to get a list of events, the external URI is

```
http://api.mydomain.com/events
```

However, internally, tomcat sees

```
http://api.mydomain.com/servlet/events
```

which causes all self links to have similar pattern. For example

```
http://api.mydomain.com/servlet/events
http://api.mydomain.com/servlet/events/:id
```

Is there an easy way to customize ControllerLinkBuilder to re-write the URIs?

nickgrealy:
Just wondering, would handling a `X-Forwarded-Path` header solve this issue? i.e.
https://github.com/nickgrealy/spring-hateoas/commit/d069038dbdc24e0f66e3d64fba57978914c9e70b

You could then automatically inject the headers as part of your application filters... e.g.
https://gist.github.com/nickgrealy/b4a582c8e1c411c5396d#file-application-java-L18-L24

I haven't committed to this project before, so let me know if I can go ahead and submit a pull request (do you want the HeaderOverrideFilter.java from the example included too?).

@aleksz - re: not working for POST - that sounds like a separate issue to this

olivergierke:
`ControllerLinkBuilder` already supports well-defined proxy server headers, so it's basically up to you making sure your proxy server sets those appropriately.

otrosien:
Should be obsolete now. Current master is delegating this to UriComponentsBuilder, which is also base-path aware. (use `X-Forwarded-Prefix` header. see also https://jira.spring.io/browse/SPR-12500)

gregturn:
Closed since we observe standard headers as well as support relative links as well.

PR #409, opened 2015-10-26
Added X-Forwarded-Path to ControllerLinkBuilder

nickgrealy:
I've added `Added X-Forwarded-Path` handling in the ControllerLinkBuilder, so that context paths can be handled.

License Agreement.

### DIFF
--- a/src/main/java/org/springframework/hateoas/mvc/ControllerLinkBuilder.java
+++ b/src/main/java/org/springframework/hateoas/mvc/ControllerLinkBuilder.java
@@ -43,6 +43,7 @@ import org.springframework.web.util.UriTemplate;
  * @author Oliver Gierke
  * @author Kamill Sokol
  * @author Greg Turnquist
+ * @author Nick Grealy
  */
 public class ControllerLinkBuilder extends LinkBuilderSupport<ControllerLinkBuilder> {
 
@@ -225,6 +226,12 @@ public class ControllerLinkBuilder extends LinkBuilderSupport<ControllerLinkBuil
 
 		if (hasText(port)) {
 			builder.port(Integer.parseInt(port));
+		}
+
+		String path = request.getHeader("X-Forwarded-Path");
+
+		if (hasText(path)){
+			builder.path(path);
 		}
 
 		return builder;
--- a/src/test/java/org/springframework/hateoas/alps/JacksonSerializationTest.java
+++ b/src/test/java/org/springframework/hateoas/alps/JacksonSerializationTest.java
@@ -70,7 +70,9 @@ public class JacksonSerializationTest {
 								).build())//
 				).build();
 
-		assertThat(mapper.writeValueAsString(alps), is(read(new ClassPathResource("reference.json", getClass()))));
+		assertThat(mapper.writeValueAsString(alps).replaceAll("\r\n", "\n"), is(read(new ClassPathResource("reference" +
+						".json", getClass())
+		)));
 	}
 
 	private static String read(Resource resource) throws IOException {
--- a/src/test/java/org/springframework/hateoas/mvc/ControllerLinkBuilderUnitTest.java
+++ b/src/test/java/org/springframework/hateoas/mvc/ControllerLinkBuilderUnitTest.java
@@ -49,6 +49,7 @@ import org.springframework.web.util.UriComponentsBuilder;
  * @author Kamill Sokol
  * @author Oemer Yildiz
  * @author Greg Turnquist
+ * @author Nick Grealy
  */
 public class ControllerLinkBuilderUnitTest extends TestUtils {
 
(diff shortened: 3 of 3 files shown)


gregturn:
I'm trying to find a spec for x-forwarded-path to check things out and I can't. Do you have link documenting proper behavior?

nickgrealy:
@gregturn - no, unfortunately I don't have any supporting documentation. 

I came across it (generally), while looking for documentation on `X-Forwarded-Host` - which [I also couldn't find in w3c.org](https://www.google.com.au/search?q="x-forwarded-host"+site:w3c.org), yet exists in `ControllerLinkBuilder`.

Unless I'm mistaken, use of HTTP header prefix `X-` [is generally discouraged](http://tools.ietf.org/html/draft-saintandre-xdash-00).

Let me know what you want me to do with this...

nickgrealy:
The issue number would probably help too... #143

nickgrealy:
@gregturn - just wondering, who can action the backlog of push requests? There's a few that have been sitting around for a while now...

olivergierke:
We can't just add some code for a header that's not specified anywhere.

thebignet:
+1  Didn't see this PR before creating #431 .

olivergierke:
As indicated above, nothing of this is going to make it unless we find a reliable source specifying the behavior of that header.

thebignet:
Apparently, the [RFC](https://tools.ietf.org/html/rfc7239) has no mention for X-Forwarded-Host nor X-Forwarded-Port.

X-Forwarded-Path is a made up header with no specification, but there seems to be a concensus as to how it should be used in proxies (see [here](https://www.npmjs.com/package/linkto), [and here](http://lists.osgeo.org/pipermail/mapproxy/2013-April/001567.html) for examples).

Since it is a non invasive solution meaning that not using this header has no side effects, and that its definition is quite clear, it can be included in spring-hateoas.

nickgrealy:
+1 for precedent... @olivergierke - https://github.com/spring-projects/spring-hateoas/commit/3f8b90e2b522f6d597804b3dcc43527c9a981f67

Is `X-Forwarded-Host` and `X-Forwarded-Port` part of a spec? and if not, should they be removed?

Thanks, Nick

gregturn:
X-Forwarded-* appears governed by a draft spec here => http://tools.ietf.org/html/draft-ietf-appsawg-http-forwarded-10

Interestingly, I found RFC7239 here => http://tools.ietf.org/html/rfc7239, which appears designed to replace all the X-Forwarded headers.

If that's not enough fun, RFC6648 deprecates the "X-" prefix.

Not sure how to square non-standard, haphazard headers with a newly minted spec for the best approach on this project.

nickgrealy:
Ok, well I'm happy with whatever you guys decide.

Just a thought, what if the `X-Forwarded-*` header handling was a togglable feature - defaulting to disabled - which people could enable (somehow)? Give the choice to the people!

sebster:
I was wondering what the specification of this seems to be.

I've run into the same issue and really need a X-Forwarded-Path or equivalent header (to a reverse proxied rest service to avoid needing cross origin requests from a html5 client). RFC7239 seems to leave room for a Forwarded-Path header in section 5.5 (Extensions), specifically mentioning the reverse proxy situation.

The problem with ControllerLinkBuilder as it is now is that there is no way to alter or extend it's behavior: the constructor is package scoped, the getBuilder() as well as the linkTo methods are all static, and the "wiring" of the UriComponentsBuilder instance is fixed by the getBuilder() method implementation. The only option I see at the moment is to just copy the class and modify it and use the modified one instead; but it's not a very nice option.

osvaldopina:
I developed a [framework](https://github.com/osvaldopina/linkbuilder) to render links in spring hateoas projects. In the current snapshot I'm working on clarifying the extension points. One of these extension points allows you to customize how url parts are obtained. 

There is a chain of reponsability in which each processing object can contribute with url parts (scheme, host, port and context path). 

[This integration test](https://github.com/osvaldopina/linkbuilder/blob/master/examples/src/test/java/com/github/osvaldopina/linkbuilder/example/extensions/requestpartsfactorylist/RequestPartsFactoryListTest.java) using [this configuration](https://github.com/osvaldopina/linkbuilder/blob/master/examples/src/main/java/com/github/osvaldopina/linkbuilder/example/extensions/requestpartsfactorylist/Application.java) shows how it can integrate your own processing object into the chain.

otrosien:
Should be obsolete now. Current master is delegating this to UriComponentsBuilder, which is base-path aware. (use `X-Forwarded-Prefix` header. see also https://jira.spring.io/browse/SPR-12500)

olivergierke:
Sure? This one is about `-Path` the ticket you linked to is about `-Prefix`.

otrosien:
Well, that should be the header to set if you want to add a base-path to your URLs.

michael-simons:
Funny, how I end up finding a ticket and personally knowing everyone involved.

The project right here is using `-Path` as well… While making the switch to "pure" Spring HATOAS without custom wrapper, we found generated URLs not working anymore. I suggested a starter containing a custom filter that maps Prefix to Path and I might just share that one.

pivotal-issuemaster:
@nickgrealy Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=409)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=409) to manually synchronize the status of this Pull Request.

nickgrealy:
Fyi @olivergierke - I'm closing this PR due to inactivity. I can reopen again if requested.

PR #411, opened 2015-10-26
Added ForwardedHeaderRequestWrapper utility class.

nickgrealy:
Added ForwardedHeaderRequestWrapper utility class - #143

License Agreement.


olivergierke:
There already is support for well defined `X-Forwarded-…` headers. So what additional, well-defined functionality does the wrapper provider? How is it supposed to be used? It's not tested at all.

gregturn:
Thanks for your effors @nickgrealy, but this is no longer relevant in light of #758.

nickgrealy:
Thanks

PR #422, opened 2016-01-07
#398 Remove unnecessary URI conversion

### LINKED ISSUE #398: ControllerLinkBuilder does not handle special query parameter characters correctly
We are experiencing issues using `ControllerLinkBuilder` with certain characters in query params, and must do some nasty hacks to get around it.

Consider this use of `ControllerLinkBuilder`:

    linkTo(methodOn(BookController.class).findBooksByTitle("Spring#\n").withRel("book"));

Note the odd goofy search string with the `#` sign and Java newline. This results in an exception rather a proper link.

The link that one might expect to be generated is:

    http://localhost/books?title=Spring#\n

Which when properly encoded would be:

    http://localhost/books?title=Spring%23%0A

While those example characters may seem strange, a valid URL is possible when encoded properly, but the link builder is instead throwing.

kevinconaway:
@olivergierke I believe this PR should fix #398 

The issue as I see is that `ControllerLinkBuilderFactory` is unnecessarily converting the `UriComponents` into a URI just to coerce it in to a `UriComponentsBuilder`

```java
UriComponents components = applyUriComponentsContributer(builder, invocation).buildAndExpand(values);
return new ControllerLinkBuilder(UriComponentsBuilder.fromUriString(components.toUriString()));
```

`UriComponentsBuilder#fromUriString` is expecting a proper URI to be passed in so any special characters that are in the URI will cause issues.  

The fix in this PR just avoids converting the `UriComponents` instance to URI string just to coerce it back to a `UriComponentsBuilder`.

### DIFF
--- a/src/main/java/org/springframework/hateoas/core/LinkBuilderSupport.java
+++ b/src/main/java/org/springframework/hateoas/core/LinkBuilderSupport.java
@@ -49,6 +49,11 @@ public abstract class LinkBuilderSupport<T extends LinkBuilder> implements LinkB
 		this.uriComponents = builder.build();
 	}
 
+	public LinkBuilderSupport(UriComponents uriComponents) {
+		Assert.notNull(uriComponents);
+		this.uriComponents = uriComponents;
+	}
+
 	/*
 	 * (non-Javadoc)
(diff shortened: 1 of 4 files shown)


olivergierke:
Thanks, Kevin! That's polished and merged.

PR #423, opened 2016-01-08
Make sure URL parts are joined correctly

### LINKED ISSUE #269: linkTo(methodOn(...)) sometimes misses slashes
I regulary use requestmapping that is partly defined on the class and on the method. This is an example:
```java
@RestController
@RequestMapping("user")
public class UserController {
    @Autowired
    BackofficeUserRepository repository;

    @RequestMapping(value = "{id}", method = RequestMethod.GET)
    public Resource<BackofficeUser> getUser(@PathVariable("id") UUID id) {
        Resource<BackofficeUser> resource = new Resource<>(repository.findUser(id));
        resource.add(linkTo(methodOn(UserController.class).getUser(id)).withSelfRel());
        return resource;
    }
}
```

This is perfectly legal spring mvc and it adds the missing slash between "user" and "{id}". But when I use the linkbuilder it doesn't do that. The result of the code above is  a link like "http://localhost/user105a8719-8d59-422d-84ba-cf53b9c93447".

As a workaround I can add the slash in my mapping, but I think the linkbuilder should come to the same conclusion as spring mvc when it comes to generating the links based in requestmapping annotations. ;-)

I'm using the version 0.16 BTW.

kevinconaway:
@olivergierke this PR should fix #269 by correctly joining the URL parts together.

I'm not sure if there is a utility in spring or spring-web to join parts together but I didn't find one.  I also didn't want to use the JDK8 Joiner because I'm fairly sure spring-hateoas needs to be backwards compatible with other JDK versions.

If there is a cleaner way to join the path segments, please let me know.

### DIFF
--- a/src/main/java/org/springframework/hateoas/core/AnnotationMappingDiscoverer.java
+++ b/src/main/java/org/springframework/hateoas/core/AnnotationMappingDiscoverer.java
@@ -109,7 +109,17 @@ public class AnnotationMappingDiscoverer implements MappingDiscoverer {
 			return typeMapping;
 		}
 
-		return typeMapping == null || "/".equals(typeMapping) ? mapping[0] : typeMapping + mapping[0];
+		return typeMapping == null || "/".equals(typeMapping) ? mapping[0] : join(typeMapping, mapping[0]);
+	}
+
+	private String join(String typeMapping, String mapping) {
+		StringBuilder builder = new StringBuilder();
+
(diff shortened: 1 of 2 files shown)


olivergierke:
I slightly changed the way the concatenation works. I now just concatenate the mappings with an intermediate slash and remove multiple slashes for the entire resulting `String`. Merged with that change.

Issue #270, opened 2014-12-09
Sample for embedding HAL Resources of different types with root resource


anywhereinfo:
Hi,
Google is full of questions around how to create embedded HAL resources of different types inside the main root resource. I looked at the Resources Interface, but that can only be leveraged by same types to model collection of a particular type.
I am currently using Spring-Hateoas to create a nested resource hierarchy and creating "_embedded" jsonproperty looks like a hack. Spring Data Rest has support for embedding entities, so it will be really helpful if you could please create a sample for embedding different type of single and collection of resources within a root resource.

krosangit:
+1

keaplogik:
+1

dminkovsky:
+1 here, too.

Here's an example of one such Google result: http://stackoverflow.com/questions/25858698/spring-hateoas-embedded-resource-support

anywhereinfo:
There is already an API to link Resource. That allows us to design relationship between Resources.
Embedding is a resource representation issue. In the stackoverflow question,  HALResource class is a representation class for a particular resource. But, i dont think that representation class should have to worry about link relationships. It should be study the resources and its relationships, but accept hints from the developer as to which linked resources qualify for embedding.

sesto:
+1

sourcekick:
+1

harttamale:
+1

jiwhiz:
+1.

This issue and #169 are my pain using Spring HATEOAS with HAL.

aglassman:
+1

keaplogik:
Ultimately this feature is not supported with Spring HATEOAS. Embedding non collection resources must be done manually if you want to support the entire hal spec. Currently our team has decided to consider this a HAL light and we do not embed non list resources. Ultimately this issue should turn into a feature request. Thoughts? May want to create a new issue requesting support.
"Spring HATEOAS to support embedding models in a resource"

aglassman:
I'd be interested in seeing a feature similar to what @anywhereinfo describes above.
> ... perhaps having Resource Representation Annotations, that can be placed on a Resource, would help to determine whether to embed Resources or not?

I'm currently using spring-data-rest, and there seems to be no easy way to specify if a child resource should appear in the _embedded object.

olivergierke:
I might be missing something but this works out of the box:

```java
// Infrastructure setup, only needed for test case
ObjectMapper mapper = new ObjectMapper();
mapper.registerModule(new Jackson2HalModule());
[...]

Issue #36, opened 2012-12-30
Hal support is incomplete


dschulten:
My experiments with the Hal support gave me this for a Resource having an attached Resources object. The Resource is Thorin, his belongings are the embedded Resources. I added a describedBy rel to each Resource and to Resources just for demonstration purposes.

```
{
    "firstname": "Thorin",
    "lastname": "Oakenshield1",
    "products": {
        "content": [
            {
                "productName": "Orcrist",
                "_links": {
                    "self": {
                        "rel": "self",
                        "href": "http://localhost:8080/products/1"
                    },
                    "describedBy": {
                        "rel": "describedBy",
                        "href": "http://example.com/doc#product"
                    }
                }
            }
        ],
        "_links": {
            "describedBy": {
                "rel": "describedBy",
                "href": "http://example.com/doc#products"
            }
        }
    },
    "_links": {
        "self": {
            "rel": "self",
            "href": "http://localhost:8080/people/1"
        },
        "describedBy": {
            "rel": "describedBy",
            "href": "http://example.com/doc#customer"
        }
    }
}
```

This is not the correct HAL representation of a resource with embedded resources. The correct representation should be:

```
{
    "firstname": "Thorin",
    "lastname": "Oakenshield1",
    "_embedded": {
        "products": [
            {
                "productName": "Orcrist",
                "_links": {
                    "self": { "href": "http://localhost:8080/products/1"},
                    "describedBy": {"href": "http://example.com/doc#product"}
                }
            }
        ],
        "_links": {
            "describedBy": {"href": "http://example.com/doc#products"}
        }
    },
    "_links": {
        "self": {"href": "http://localhost:8080/people/1"},
        "describedBy": {"href": "http://example.com/doc#customer"}
    }
}
```

For decent HAL support we would also need templated URIs, for something like:

```
"find": { "href": "/orders{?id}", "templated": true }
```

harttamale:
+1

TomStandaert:
+1

jiwhiz:
I got same result like @tombee. @olivergierke @dschulten can we re-open this issue?

pivovarit:
I am starting working on this issue on my own. Let's see what I can do.

elnur:
@pivovarit good luck! :+1:

Issue #175, opened 2014-05-05
Name embedded collections


drdamour:
The Resources type (and any derived types, ie PagedResource) should allow you to name the collection that will eventually be the embedded set.  It's auto named now, but the api developer should really have control over this.

gjrwebber:
+1

marvinrichter:
For those who didn't get Olivers answer straight away, like me:

If you want the plural form of the class' name rather than uncapitalized class.getSimpleName + "List", you just need to use the EvoInflectorRelProvider instead of the standard DefaultRelProvider.

If you use Spring Boot, just add the following dependency
```xml
<dependency>
    <groupId>org.atteo</groupId>
    <artifactId>evo-inflector</artifactId>
    <version>1.2</version>
</dependency>
```

otherwise you need to also add the following bean to your configuration class
```java
@Bean
public RelProvider relProvider() {
    return new EvoInflectorRelProvider();
}
```

These results in returning
```json
"_embedded": {
    "users": [...]
}
```
instead of
```json
"_embedded": {
    "userList": [...]
}
```

olivergierke:
As documented in the [reference documentation](http://docs.spring.io/spring-hateoas/docs/current/reference/html/#spis.rel-provider). :smile:

marvinrichter:
lol. I totally missed the point 3. 😄

Jeckyli:
+1. 

Also add to OP's comment, what's the rationale to have PagedResources returned under _embedded with class name? Is it possible to return just the array of data like non-HAL format? 

Like for ```/v1/users/``` just return

```JSON
"_embedded": 
[
  {
    "name": "Marvin"
  },
  {
    "name": "Alex"
  }
]
```

This way UI doesn't have to know the property name to look for under _embedded.

olivergierke:
The [HAL specification](https://tools.ietf.org/html/draft-kelly-json-hal-06#section-4.1.2).

stephaneeybert:
Even if we can have, for example the name 'users' instead of the name 'userResourceList', we still have a name varying for each resource type. It'd be helpful to configure a static name, common to all resource types, like 'items' for example.

marvinrichter:
@stephaneeybert having one static name for all collections would be very easy with the solution that @olivergierke suggested first.

[...]

PR #428, opened 2016-01-22
support for embedding HAL resources. serialization only.

trombka:
See #270 and #36

### DIFF
--- a/src/main/java/org/springframework/hateoas/ResourceSupport.java
+++ b/src/main/java/org/springframework/hateoas/ResourceSupport.java
@@ -17,14 +17,18 @@ package org.springframework.hateoas;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 
 import javax.xml.bind.annotation.XmlElement;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import org.springframework.util.Assert;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import org.springframework.util.CollectionUtils;
 
 /**
  * Base class for DTOs to collect links.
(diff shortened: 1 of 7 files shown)


src/main/java/org/springframework/hateoas/ResourceSupport.java
```diff
@@ -58,14 +65,21 @@ public void add(Link link) {
...
+	 * Adds all given {@link Link}s or {@link EmbeddedResource}s to the resource.
 	 * 
-	 * @param links
+	 * @param linksOrEmdeddedResource
 	 */
-	public void add(Iterable<Link> links) {
-		Assert.notNull(links, "Given links must not be null!");
-		for (Link candidate : links) {
-			add(candidate);
+	public void add(Iterable linksOrEmdeddedResource) {
```
karlthepagan: Is there a compelling reason to avoid strong typing here?

trombka: You cannot have both add(Iterable<Link>) and add(Iterable<EmbeddedResource>) as explained here: http://stackoverflow.com/questions/1998544/method-has-the-same-erasure-as-another-method-in-type
Leaving all `add` methods for `Link` and having `addEmbeddeResource` would kind of break a naming convention.
In my opinion it would be better to deprecate `add` in favour of `addLink(s)` and `addEmbeddedResource(s)`. However this would be a change of a public interface of the class, so I thought it was not suitable for a simple PR.

karlthepagan: IMO addEmbedded or addEmbeddedResource is a justified addition to the API.

pivotal-issuemaster:
@trombka Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=428) to manually synchronize the status of this Pull Request.

gregturn:
See https://github.com/spring-projects/spring-hateoas/issues/270#issuecomment-470976112 and https://github.com/spring-projects/spring-hateoas/issues/175#issuecomment-470974583 for examples of a fluent API I'm working on to simplify building `RepresentationModel` objects.

NOTE: I still don't have a way to mix both top level data attributes AND embedded stuff, which is clearly denoted in the spec. That may require something more fundamental.

PR #431, opened 2016-02-02
Added support for X-Forwarded-Path header

thebignet:
Some proxies use URL rewriting, which means that requested URI and forwarded URI do not match.  In such cases, URLs generated by ControllerLinkBuilder are incorrect.

The following issue suggest usage of a X-Forwarded-Path header which would hold the base path of the service.
https://github.com/18F/api.data.gov/issues/260

The path method of the builder was used with the X-Forwarded-Path header value.


thebignet:
Discovered that it's actually a duplicate of #409, which is marked as unlikely.  This behaviour is really missing in spring-hateoas and in the RFC and is due to URL rewriting.

olivergierke:
As indicated above, nothing of this is going to make it unless we find a reliable source specifying the behavior of that header.

olivergierke:
Duplicates #409.

Issue #438, opened 2016-03-09
Compatibility with Jackson 2.7


sylvainlaurent:
Spring HATEOAS 0.19.0 is not compatible with Jackson 2.7.2 :

> Caused by: java.lang.NoSuchMethodError: com.fasterxml.jackson.databind.deser.std.ContainerDeserializerBase: method <init>(Ljava/lang/Class;)V not found
	at org.springframework.hateoas.hal.Jackson2HalModule$HalLinkListDeserializer.<init>(Jackson2HalModule.java:519)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:408)
	at org.springframework.beans.BeanUtils.instantiateClass(BeanUtils.java:147)

It seems that the constructor that was removed was deprecated since Jackson 2.3...
see https://github.com/FasterXML/jackson-databind/commit/a5301ba825a33f95ca9749655d2ccf67c0f512a3#diff-431839367c5eccdcd048960395c0782d

gregturn:
@sylvainlaurent Have you tested this patch against both Jackson 2.6 and 2.7? Backwards compatibility is a real bug bear with Jackson and something to watch out for.

sylvainlaurent:
seems to be OK with both as far as I could test

philwebb:
Any chance this can be considered soon? It's currently a blocker for Boot 1.4

MHarris021:
+1

olivergierke:
Duplicated by #455.

PR #439, opened 2016-03-11
Fix Compatibility with Jackson 2.7 #438

sthibert:
Here is a suggestion to fix issue #438.
This solution has been tested and works fine for me.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -60,7 +60,7 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<spring.version>4.1.7.RELEASE</spring.version>
 		<logback.version>1.1.3</logback.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -60,7 +60,7 @@
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<spring.version>4.1.7.RELEASE</spring.version>
 		<logback.version>1.1.3</logback.version>
-		<jackson.version>2.5.4</jackson.version>
+		<jackson.version>2.7.2</jackson.version>
```
sylvainlaurent: I would stick to 2.6.5 since it is the version supported by spring boot 1.3.3.

olivergierke:
Already fixed with the patch in #455.

PR #442, opened 2016-03-28
Fix embedded monospacing in documentation

davidmorley:
Monospaced quoting was rendering incorrectly in the `EntityLinks` section because alphabetic characters are not considered a valid boundary for constrained quotes. I switched it to use an unconstrained quote which does not have the same constraint. 

See: http://www.methods.co.nz/asciidoc/userguide.html#X52


gregturn:
Superceded by 078e1126c93aaaf8fc15185d2ac825164dbc2cfb

PR #447, opened 2016-04-21
Support form creation

gillarramendi:
We have been reviewing the state of the art regarding HATEOAS within Java environments such as Spring HATEOAS and JAX-RS and none of the implementations include any form (create/update) support, only link creation.

Taking into account that the most important hypermedia format (HTML) includes a great **support for forms**, offering a complete and self-contained description of the update/create options offered by a web app, we consider that it would be an interesting feature to **add this support to Spring HATEOAS**.
It means that clients will know which is the contract offered by the API update/create interactions, including the **possible fields** and if necessary the **possible values** for each property/field. In other words, the hypermedia format will guide the developer to implement form interactions with a self-contained format.

Of course, the first decision to add this support is to define the hypermedia format. There are already some hypermedia formats that support forms, like Hydra and Siren. We consider that the most natural option that fits with Spring HATEOAS, based on HAL, is **HAL-FORM** (https://github.com/mamund/hal-forms). Basically, an optional extension of HAL for forms support.

As an starting point, we created a first draft implementation including a **form builder** that allows the definition of the form fields and a *_MessageConverter *_that renders it in HAL-FORM format.  In addition, we have proposed some updates within HAL-FORM specification, basically to determine the possible values for each field, and we are in conversations with Mike Amundsen @mamund, the specification author, to include them in the specification.
- https://github.com/mamund/hal-forms/issues/9
- https://github.com/mamund/hal-forms/issues/10

We created an **example application** that shows the usage of the new API to create Forms (https://github.com/hdiv/hal-forms-sample).

For example, a **form with two fields** is created in this way:

``` java
ControllerFormBuilder formBuilder = ControllerFormBuilder
    .formTo(ControllerFormBuilder.methodOn(TaskController.class).create(new Task()));

formBuilder.property("description").readonly(false);
formBuilder.property("priority").suggest().values(new Priority[]{Priority.High, Priority.Low});
Form form = formBuilder.withDefaultKey();

resource.add(form);
```

And the generated **HAL-FORMS response**:

``` json
{
  "_links": {
    "self": {
      "href": "http://localhost:8080/tasks"
    }
  },
  "_templates": {
    "default": {
      "method": "POST",
      "properties": [
        {
          "name": "description",
          "readOnly": false
        },
        {
          "name": "priority",
          "suggest": {
            "values": [
                "HIGH",
                "MEDIUM",
                "LOW"
              ]
          }
        }
      ]
    }
  }
}
```

Any thoughts will be appreciated.
Thanks in advance.

@robertovelasco, @iunanua, @gillarramendi


olivergierke:
This looks pretty decent. Would you mind squashing the commits into a single one and summarizing the changes in a commit message? We might wanna move all the model types to create the forms into the `hal` sub package as I think we should make clear which of the types are specifically available to build media-type specific representations.

Also, I'd like to check with @dschulten for some overlap in the work he has done in his Spring HATEOAS extension to support Siren and Hydra which also produce additional metadata in representations. We were already thinking of bringing those extensions back to Spring HATEOAS at some point and I'd like to make sure we don't create totally different ways to declare metadata for the various formats. Might be a good chance to pick up those efforts again.

@gillarramendi — I've seen that a [team mate](http://www.springio.net/speaker/roberto-velasco/) of you will be speaking at Spring I/O in Barcelona. Will you be around? Might be a good chance to talk and discuss how to proceed with this one.

gillarramendi:
Thanks for the feedback!
Ok, we will create another pull request including all changes into one commit and move all classes to `hal.forms` package.

Totally agree with the creation of a common metadata, shared between different format implementations. We are very interested in collaborating in this topic with @dschulten to pull in the same direction.

Yes, I'll be in the Spring I/O in Barcelona, we can talk about it there.

olivergierke:
It should be fine if you just squash the commits and force them to the branch you created the PR from. Looks you didn't create a feature branch of some sorts so it might indeed be tricky. Just do whatever you think is fine, a new PR works well for me.

Looking forward to see you in Barcelona! 🙃

dschulten:
Hi,

Am 28.04.2016 um 18:48 schrieb Gotzon Illarramendi:
> Thanks for the feedback!
> Ok, we will create another pull request including all changes into one
> commit and move all classes to |hal.forms| package.
> 
> Totally agree with the creation of a common metadata, shared between
> different format implementations. We are very interested in
> collaborating in this topic with @dschulten
> <https://github.com/dschulten> to pull in the same direction.

Manual form creation and reflective form creation complement each other
nicely. I think one really needs both, just like links - you can create
links based on reflection *or* manually today.

Common interfaces are the solution, I think.
There could be one implementation that has reflection data inside, and
another one with "manually built" form information.

Affordance could be the common ground. It is already meant as generic
request metadata descriptor. I have created and released new
ActionDescriptor and ActionInputParameter interfaces to Maven central
just now from the prior implementation classes, they now abstract away
the underlying reflection-based impl.

Short reasoning why these interfaces are needed in the current form:

I think of a form descriptor (hypermedia affordance) as a structure
having three levels.

1. The first level is the resource URI or URI template itself. That is
already covered by Link and also by Affordance, which inherits from it.
Affordance enhances Link in that it contains all features of a http Link
and Link-Template header, its toString method gives you a link header
you can add to outgoing responses if the response mediatype has no link
construct.

2. On the URI of the affordance you can have several possible methods,
that is why one Affordance has multiple ActionDescriptors.

Action descriptors are the second level of the structure.
Each action descriptor represents a http method execution, with things
like the http method, an identifier for the action (think http form
name), an optional semantic action type (think
http://schema.org/SearchAction, which is already widely used on the Web).

3. Furthermore each http method execution involves headers and a body.

That brings us to the third level: body and request header descriptors.
The headers are key-value pairs, the expected body is a potentially
nested data structure. That is what ActionInputParameter is about.

Please have a look if you can bring your metadata behind the Affordance
impl using the latest ActionDescriptor and ActionInputParameter
interfaces from spring-hateoas-ext 0.3.0 beta1. Would that be possible?
If you need any changes on the interfaces, we can adjust quite easily, I
think.

Best regards,
Dietrich

olivergierke:
That's great feedback, Dietrich! Do you have a few pointers into where we should start looking in the extensions you have? I'd be interested to see (well, I can google those things I guess, but I'd like to have them listed here for reference):

1. A pointer to your affordance API.
2. Some code actually using it (is there a tiny sample maybe?).
3. 2 would be even better if it was an executable Boot project or the like so that I could start playing with your API and the HAL forms support to see how much they look similar or just in general explore the approaches.

dschulten:
# Pointers to the API

* [Affordance](https://github.com/dschulten/hydra-java/blob/master/spring-hateoas-ext/src/main/java/de/escalon/hypermedia/affordance/Affordance.java) class

Users can create it manually ilke `Link`, or use `AffordanceBuilder.linkTo/methodOn` methods.

`AffordanceBuilder` allows to add multiple pointers to action handlers using `.and`, see [example](https://github.com/dschulten/hydra-java#affordancebuilder-for-rich-hyperlinks).

* [ActionDescriptor](https://github.com/dschulten/hydra-java/blob/master/spring-hateoas-ext/src/main/java/de/escalon/hypermedia/affordance/ActionDescriptor.java) interface is returned from Affordance.getActionDescriptors and describes method executions

* [ActionInputParameter](https://github.com/dschulten/hydra-java/blob/master/spring-hateoas-ext/src/main/java/de/escalon/hypermedia/affordance/ActionInputParameter.java) interface is returned from ActionDescriptor.getActionInputParameter(String name). The input parameter names can be obtained from ActionDescriptor.getXXXNames.
If ActionDescriptor.hasRequestBody is true, then getRequestBody gives you the descriptor for the request body.

# Sample application

Simple webapp ([war](https://github.com/dschulten/hydra-java/tree/master/hydra-sample)). I'll spring-bootify it to make it easier to work with.

It is also deployed on [OpenShift](http://jbosswildfly-escalon.rhcloud.com/hypermedia-api/events), hit F5 several times if the container is not running. (Not great, Openshift!) 
You get XHtml in the browser at first, from the XhtmlMessageConverter. Use Postman to Accept application/hal+json, application/vnd.siren+json, application/ld+sjon

gillarramendi:
Thank you @dschulten for your great feedback.

We agree that form creation based on annotations and using a Java API are complementary and the developers can choose the best solution for their projects. Both systems can work with a common metadata and Affordance looks good for that.

Regarding the implementation of Java API based approach, we see two possible approaches:
  - Specific API, tied to the final format: The FormBuilder will expose methods to populate metadata only for a specific hypermedia format. For example forms in `hal-forms` are defined in the `template` elements, so the API have an `addTemplate(..)` method.

Another example to define possible values for a field:
```java
HalFormBuilder formBuilder = HalFormBuilder
	.formTo(ControllerFormBuilder.methodOn(TaskController.class).create(new Task()));
formBuilder.property("category").suggest().embedded(categories.findAll());
```
Where `property`, `suggest` and `embedded` are `hal-form` specific attributes.



  - Generic API that works for all formats. The same example above.

```java
AffordanceBuilder formBuilder = AffordanceBuilder
	.formTo(ControllerFormBuilder.methodOn(TaskController.class).create(new Task()));
formBuilder.parameter("category").possibleValues(categories.findAll());
```
Here the problem is that some metadata you can add to the Affordance won't be used by the message converter. For example, possible values for a field are shown in `hal-forms` but not in `hydra`.

As far as we know this is the approach used by the current affordance implementation based on Annotations.

What do you think about these two options?


Regarding the source code location, we agree to take @dschulten code as foundation (`hydra-java` project) for the final pull request and if you agree we can provide the following elements to your code making pull request there:

- Java API for the Affordance creation
- Message converter for hal-forms
- Sample application using hal-forms

Another additional task we see necessary in order to integrate everything within `spring-hateoas`, that at least from our side would be the goal, is to adapt the `spring-hateoas-ext` project to `spring-hateoas`, in order to prepare a PR. Basically package changes and some code refactor are needed.

dschulten:
The distinction between resource and representation gives us some guidance here I think.

What we are doing in the controller code is to describe the resource state. We say, the *resource* has specific hypermedia affordances; based on the incoming request and the persisted states of the server's resources, the server chooses the available hypermedia affordances.

It is the responsibility of the message converter to *represent* the resource to the client, based on content negotiation. At least that is how Spring has separated the concerns.

An example why this separation makes sense: If we code to one particular representation (media-type), we would have to code the same form description again for a second media type, because it uses different naming and structure. What does the first message converter do with the metadata that aren't meant for it? Or do we have to look at the Accept header in the controller code to decide which form api to use?

OTOH, if we go for the common descriptor object, you are correct that it will have to be quite comprehensive. Some media-types won't make use of some of the information. Why do you see this as problematic? The opposite might be a problem, if a media type would force us to describe something which is really specific to that *media-type* and not to characteristics of the resource. For the reflection-based approach, a converter jar with matching annotations might give people the necessary flexibilty (I have done that with the @Expose annotation, which comes with the hydra converter). An API-based user will probably have to customize the form builder and add that customization to the converter package.

For the code location, I would like to suggest that we commit to the feature/affordances branch in this repo. @olivergierke what do you think?

What I have on my list is to get the latest changes to spring-hateoas ext into the feature branch this weekend.

olivergierke:
The topic of media types specific configuration or definitions has come up for the very core types already. I really think that we should try to capture the essence of affordances in media type agnostic APIs but then allow plugging in media type specific declarations. While this is kind of hard to achieve on core types, I think the affordance builder could offer something like this:

```java
….withMediaTypeSpecificDetails(forHal().halSpecificInformation());
….withMediaTypeSpecificDetails(forJsonLd().jsonLdSpecificInformation());
```

The individual `HttpMessageConverter`s could then make sure that the types produced by those media type specific calls get marshaled properly.

olivergierke:
@dschulten — I just checked and it looks like an can rebase your feature branch onto current master easily. I'd like to update it to that before you go ahead and port the latest changes. Is that okay for you?

dschulten:
Very much ok!

olivergierke:
Feature branch is rebased. Looks like we introdouced a new method in the `LinkBuilder` API that `AffordanceBuilder` will have to implement but that one should be straight forward to implement.

gillarramendi:
ok, we will create as generic as possible API that will cover common cases and a 'plugin' system to add media-type specific cases.

gregturn:
Superceded by #612 and resolved via https://github.com/spring-projects/spring-hateoas/commit/70448a854098343897fd99e3227b4a61db54b921.

Final solution isn't precisely the same thing, but we plan to iterate over it in the near future and expand to support other mediatypes. Feel free to check the HAL-FORMS section of code and their corresponding unit tests.

Also, see monitor [Spring HATEOAS Examples repo](https://github.com/spring-projects/spring-hateoas-examples) as we plan to add an Affordances example very soon.

pivotal-issuemaster:
@gillarramendi Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=447)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=447) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@gillarramendi Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=447)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=447) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@gillarramendi Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=447)!

Issue #465, opened 2016-07-15
ControllerLinkBuilder appends default port for http/https scheme


igorbolic:
`ControllerLinkBuilder` seems to handle differently `X-Forwarded-Host` header, then the `ServletUriComponentsBuilder`. 

The `getBuilder()` method uses `ServletUriComponentsBuilder` i.e. `UriComponentsBuilder` to generate controller links. While the `UriComponentsBuilder` ignores default ports _80/443_ for the protocols _http/https_, the `ControllerLinkBuilder` does not do the same when the **X-Forwarded-Host** header is detected:

```
String host = forwarded.getHost();
host = hasText(host) ? host : request.getHeader("X-Forwarded-Host");

if (!hasText(host)) {
    return builder;
}
```

The header be handled in the same way as in the `UriComponentsBuilder` i.e. to use already initialized `host` & `port` values from the builder. Made a pull request https://github.com/spring-projects/spring-hateoas/pull/466

PR #466, opened 2016-07-15
Remove processing of X-Forwared-Host header

igorbolic:
When the X-Forwarded-Host is sent with the default protocole port e.g.
somethingDifferent:80, the port is added to the link although it was
already processed by ServletUriComponentsBuilder which ignores ports
80/443 for HTTP/HTTPS. The processing of X-Forwarded-Host has been removed;
Now the default port is not set in the link.

License Agreement.


otrosien:
@igorbolic handling of X-Forwarded-Host is delegated to UriComponentsBuilder () from spring framework now. AFAICS this PR and issue #465 can be closed. You might want to try the current BUILD-SNAPSHOT.

PR #470, opened 2016-07-26
Add HAL_JSON_UTF8 to MediaTypes

drumonii:
Add `;charset=UTF-8` to `application/hal+json` similar to `APPLICATION_JSON_UTF8` in `MediaType`

### DIFF
--- a/src/main/java/org/springframework/hateoas/MediaTypes.java
+++ b/src/main/java/org/springframework/hateoas/MediaTypes.java
@@ -22,6 +22,7 @@ import org.springframework.http.MediaType;
  * 
  * @author Oliver Gierke
  * @author Przemek Nowak
+ * @author Drummond Dawson
  */
 public class MediaTypes {
 
@@ -34,4 +35,14 @@ public class MediaTypes {
 	 * Public constant media type for {@code application/hal+json}.
 	 */
 	public static final MediaType HAL_JSON = MediaType.valueOf(HAL_JSON_VALUE);
+
+	/**
+	 * A String equivalent of {@link MediaTypes#HAL_JSON_UTF8}.
+	 */
+	public static final String HAL_JSON_UTF8_VALUE = HAL_JSON_VALUE + ";charset=UTF-8";
+
+	/**
+	 * Public constant media type for {@code application/hal+json;charset=UTF-8}.
+	 */
+	public static final MediaType HAL_JSON_UTF8 = MediaType.valueOf(HAL_JSON_UTF8_VALUE);
 }
--- a/src/main/java/org/springframework/hateoas/core/JsonPathLinkDiscoverer.java
+++ b/src/main/java/org/springframework/hateoas/core/JsonPathLinkDiscoverer.java
@@ -67,23 +67,22 @@ public class JsonPathLinkDiscoverer implements LinkDiscoverer {
 	}
 
 	private final String pathTemplate;
-	private final MediaType mediaType;
+	private final List<MediaType> mediaTypes;
 
 	/**
 	 * Creates a new {@link JsonPathLinkDiscoverer} using the given path template supporting the given {@link MediaType}.
 	 * The template has to contain a single {@code %s} placeholder which will be replaced by the relation type.
 	 * 
 	 * @param pathTemplate must not be {@literal null} or empty and contain a single placeholder.
-	 * @param mediaType the {@link MediaType} to support.
+	 * @param mediaTypes the {@link MediaType}s to support.
 	 */
-	public JsonPathLinkDiscoverer(String pathTemplate, MediaType mediaType) {
-
+	public JsonPathLinkDiscoverer(String pathTemplate, MediaType... mediaTypes) {
 		Assert.hasText(pathTemplate, "Path template must not be null!");
 		Assert.isTrue(StringUtils.countOccurrencesOf(pathTemplate, "%s") == 1,
 				"Path template must contain a single placeholder!");
 
 		this.pathTemplate = pathTemplate;
-		this.mediaType = mediaType;
+		this.mediaTypes = Arrays.asList(mediaTypes);
 	}
 
 	/* 
@@ -178,6 +177,16 @@ public class JsonPathLinkDiscoverer implements LinkDiscoverer {
 	 */
 	@Override
 	public boolean supports(MediaType delimiter) {
-		return this.mediaType == null ? true : this.mediaType.isCompatibleWith(delimiter);
+		boolean supports = false;
+		if (this.mediaTypes == null) {
+			return true;
+		} else {
+			for (MediaType mediaType : this.mediaTypes) {
+				if (mediaType.isCompatibleWith(delimiter)) {
+					supports = true;
+				}
+			}
+		}
+		return supports;
 	}
 }
--- a/src/test/java/org/springframework/hateoas/mvc/TypeReferencesIntegrationTest.java
+++ b/src/test/java/org/springframework/hateoas/mvc/TypeReferencesIntegrationTest.java
@@ -17,6 +17,7 @@ package org.springframework.hateoas.mvc;
 
 import static org.hamcrest.Matchers.*;
 import static org.junit.Assert.*;
+import static org.springframework.test.web.client.ExpectedCount.times;
 import static org.springframework.test.web.client.MockRestServiceServer.*;
 import static org.springframework.test.web.client.match.MockRestRequestMatchers.*;
 import static org.springframework.test.web.client.response.MockRestResponseCreators.*;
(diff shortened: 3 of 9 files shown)


src/main/java/org/springframework/hateoas/MediaTypes.java
```diff
@@ -34,4 +35,14 @@
 	 * Public constant media type for {@code application/hal+json}.
 	 */
 	public static final MediaType HAL_JSON = MediaType.valueOf(HAL_JSON_VALUE);
+	
+	/**
+	 * A String equivalent of {@link MediaTypes#HAL_JSON_UTF8}.
+	 */
+	public static final String HAL_JSON_UTF8_VALUE = HAL_JSON_VALUE + ";charset=UTF-8";
```
eneveu: Instead of doing String concatenation, I would write `= "application/hal+json;charset=UTF-8"`, to make this easier to grep and to understand. It would be similar to `org.springframework.http.MediaType`.

Also, I think we should add an underscore to "UTF8" in the constant names: HAL_JSON_UTF_8_VALUE. It's more readable, and it seems to be more conventional when looking at the Spring / Guava / JDK codebases.

drumonii: I agree about the String concatenation, but it's interesting you mention `MediaType` because that's where I modeled my `hal+json` additions from. See [this line](https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/java/org/springframework/http/MediaType.java#L103)

Also it uses `UTF8` instead of `UTF_8` as you suggested. Just want to be consistent how the current framework does it.

eneveu: Indeed, I missed that "APPLICATION_JSON_UTF8_VALUE" constant in MediaType... It's weird that they chose to write it like that, since Guava / Spring / JDK seem to prefer the underscore everywhere else (I even cloned a few Spring project and grep'ed to check :D ). 

I guess it makes sense in a way, since you clearly separate the "HAL" / "JSON" and "UTF 8" info. We should probably go for consistency with MediaTypes.

Anyway I created a temporary class "AdditionalMediaTypes" in my project, until we can get this in Spring HATEOAS. Great suggestion :)

drumonii: Thanks! Yeah, I had to do the same to have a `hal+json` media type with `;charset=UTF-8`. Hopefully this gets in the next release.

I agree it's an odd way to write it but in this situation I'd rather keep it as Spring's `MediaType` does rather than be different only here.

marceloverdijk:
This would be a nice addition especially as a default Boot app with the hateoas starter adds the charset to the content type.

DevWantJustHaveFun:
+1 for this addition

gregturn:
It appears that the Spring Framework has APPLICATION_JSON_UTF8 only to support the fact that `MappingJackson2HttpMessageConverter` defaults to `application/json ;charset=UTF-8`. In other words, in that same class, there are no other UTF-8 types added to the vast list of IANA spec'd mediatypes.

So my question is, why do we need to add one for HAL in this project? Is something breaking because of it? Can you draft a test case or scenario where this is causing issues? Otherwise, most of the dialog has been about how to implement the solution to a problem not well defined.

drumonii:
Spring Boot defaults its HTTP encoding to `UTF-8` since 1.2, see [here](https://github.com/spring-projects/spring-boot/commit/109c3a343963ed4c861d775ce571d49c5b0639f2).

The issue that I came across is that if you had a test like this, it would fail.

```
mockMvc.perform(get("/api/test"))
		...
		.andExpect(content().contentType(MediaTypes.HAL_JSON))

java.lang.AssertionError: Content type 
Expected :application/hal+json
Actual   :application/hal+json;charset=UTF-8
```

Since there is a `UTF-8` version of `application/json` in Spring Framework's `MediaType`, thought it would be useful if SDR had the `UTF-8` version of `application/hal+json` in its `MediaTypes` for consistency.

src/main/java/org/springframework/hateoas/MediaTypes.java
gregturn: Based upon Spring's MediaType being:

```
/**
 * Public constant media type for {@code application/json;charset=UTF-8}.
 */
public final static MediaType APPLICATION_JSON_UTF8;

/**
 * A String equivalent of {@link MediaType#APPLICATION_JSON_UTF8}.
 */
public final static String APPLICATION_JSON_UTF8_VALUE = APPLICATION_JSON_VALUE + ";charset=UTF-8";
```
...I think the proposed fix is crafted properly.

gregturn:
Additionally, we need some tests for this mediatype such as in `EnableHypermediaSupportIntegrationTest` and `TypeReferencesIntegrationTest`.

In fact, if you blanket check for usage of the media type, there are other classes that must be revised based on such a change including: `HypermediaSupportBeanDefinitionRegistrar` and `HalLinkDiscovered`.

Updating these spots in this PR would go a long way toward polishing up this PR for merging.

drumonii:
@gregturn I'll take a look at updating the tests. Will get back to you in Spring Data Gitter if I have questions.

gregturn:
Thanks. FYI @'ing me here is more effective to reach me than gitter. :)

src/main/java/org/springframework/hateoas/core/JsonPathLinkDiscoverer.java
```diff
@@ -67,23 +67,22 @@
...
 	 * Creates a new {@link JsonPathLinkDiscoverer} using the given path template supporting the given {@link MediaType}.
 	 * The template has to contain a single {@code %s} placeholder which will be replaced by the relation type.
 	 * 
 	 * @param pathTemplate must not be {@literal null} or empty and contain a single placeholder.
-	 * @param mediaType the {@link MediaType} to support.
+	 * @param mediaTypes the {@link MediaType}s to support.
 	 */
-	public JsonPathLinkDiscoverer(String pathTemplate, MediaType mediaType) {
-
+	public JsonPathLinkDiscoverer(String pathTemplate, MediaType... mediaTypes) {
```
drumonii: Wasn't quite sure what to do with this constructor. It seems that the `MediaType` is nullable so I made it varargs. I thought about another constructor with a `List` of `MediaType`...

src/main/java/org/springframework/hateoas/core/JsonPathLinkDiscoverer.java
```diff
@@ -178,6 +177,16 @@ private JsonPath getExpression(String rel) {
 	 */
 	@Override
 	public boolean supports(MediaType delimiter) {
-		return this.mediaType == null ? true : this.mediaType.isCompatibleWith(delimiter);
+		boolean supports = false;
+		if (this.mediaTypes == null) {
+			return true;
+		} else {
+			for (MediaType mediaType : this.mediaTypes) {
```
drumonii: Wasn't sure what to do here either with the `List` of `MediaType` and `isCompatibleWith`

src/test/java/org/springframework/hateoas/mvc/TypeReferencesIntegrationTest.java
```diff
@@ -82,12 +83,17 @@ public void setUp() {
 	@Test
 	public void usesResourceTypeReference() {
 
-		server.expect(requestTo("/resource")).andRespond(withSuccess(RESOURCE, MediaTypes.HAL_JSON));
+		server.expect(times(1), requestTo("/resource")).andRespond(withSuccess(RESOURCE, MediaTypes.HAL_JSON));
+		server.expect(times(1), requestTo("/resource")).andRespond(withSuccess(RESOURCE, MediaTypes.HAL_JSON_UTF8));
```
drumonii: Duplicated server expect with `MediaTypes.HAL_JSON_UTF8`. Not sure the preferred way to test the `MediaTypes.HAL_JSON` in addition to `MediaTypes.HAL_JSON_UTF8`

drumonii:
@gregturn initial polish of PR with tests is up! See my comments on unsure spots. Let me know what changes or additions elsewhere should be made. Thanks.

gregturn:
@drumonii With all the work in Berlin this week, I can hopefully take a close look next week.

drumonii:
@gregturn No worries, that sounds good!

DevWantJustHaveFun:
In what version of hateos this fix will be available ?

gregturn:
I honestly don't know. Outcome of Berlin was to work on Affordances. That took weeks and based on the outcome we are taking a different approach. I'm presently coding HAL-Forms media type and using all my lessons learned on Affordances to write a lighter version of such an API. That along with all the work on Spring Data 2 has barred Oliver from reviewing the other PRs on this project.

olivergierke:
@drumonii — Would you mind to provide us with your full name and a proper email address?

olivergierke:
Actually, nevermind. I'll have to roll back most of the changes you made anyway except the addition of the addition of the media types.

Issue #16, opened 2012-09-21
Resources should match the Collection+JSON convention


rgladwell:
Recently came across this syntax for collections of JSON object:

http://www.amundsen.com/media-types/collection/

and wondered if the Resources<T> generic type in spring-hateoas should meet it? Discussion?

gregturn:
Superceded by #482

PR #482, opened 2016-08-20
Add support for Collection+JSON media type

gregturn:


gregturn:
@olivergierke I fixed all the issues such that is works for current, spring42, spring43, spring43-next, and spring5.

gregturn:
@olivergierke Would you mind reviewing this PR to see if we can add JSON Collection support?

gregturn:
Related issue: #16

gregturn:
@olivergierke I believe I've hammered everything out.

* Complete suite of unit and integration tests including a Spring MVC REST controller using affordances, that both support HAL-FORMS and Collection+JSON, verifying they interoperate in each other's presence.
* Verification of spec fragments of JSON from http://amundsen.com/media-types/collection/examples/ in another unit test
* An interactive web test where it creates a new object then queries the newly created object basd on location header.
* Polished everything to have proper formatting.
* Also used as much package-private visibility as possible, to limit surface of the API.
* Refactored bits of HAL-FORMS to leverage the property-based utilities I added for Collection+JSON. This includes support for Jackson Ignore annotations in *both* Collection+JSON and HAL-FORMS.

So, if we can get this reviewed to push it forward, that would be great!

PR #484, opened 2016-08-22
Fixes #483 - The asterisk in URI Template donates a composite value

### LINKED ISSUE #483: UriTemplate Does not handle templates ending with an asterisk
We're connecting to a service which returns URI's in the following form:
`http://someserver.net/models?makeId=bbb9fdae593d5ce834a979f3c5f0437c{&id*,aPacks*,dataIssue}`

The regex in the UriTemplate cannot handle the '*' at the end of the template name.

`private static final Pattern VARIABLE_REGEX = Pattern.compile("\\{([\\?\\&#/]?)([\\w\\,]+)\\}");`

Can this regex get changed to:

`private static final Pattern VARIABLE_REGEX = Pattern.compile("\\{([\\?\\&#/]?)([\\w\\,*]+)\\}");`

?

JamesERi:
The URITemplate needs to handle composite values which are indicated with an asterisk.  A composite field can be expanded to a list or an associative array. https://tools.ietf.org/html/rfc6570#section-2.4.2

### DIFF
--- a/src/main/java/org/springframework/hateoas/UriTemplate.java
+++ b/src/main/java/org/springframework/hateoas/UriTemplate.java
@@ -22,6 +22,7 @@ import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -40,7 +41,7 @@ import org.springframework.web.util.UriComponentsBuilder;
  */
 public class UriTemplate implements Iterable<TemplateVariable>, Serializable {
 
-	private static final Pattern VARIABLE_REGEX = Pattern.compile("\\{([\\?\\&#/]?)([\\w\\,]+)\\}");
+	private static final Pattern VARIABLE_REGEX = Pattern.compile("\\{([\\?\\&#/]?)([\\w\\,*]+)\\}");
 	private static final long serialVersionUID = -1007874653930162262L;
 
 	private final TemplateVariables variables;;
@@ -67,7 +68,14 @@ public class UriTemplate implements Iterable<TemplateVariable>, Serializable {
 			String[] names = matcher.group(2).split(",");
 
 			for (String name : names) {
-				TemplateVariable variable = new TemplateVariable(name, type);
+				boolean isCompositeValue = name.endsWith("*");
+				TemplateVariable variable;
+
+				if (isCompositeValue) {
+					variable = new TemplateVariable(name.substring(0, name.length() - 1), VariableType.COMPOSITE_PARAM);
+				} else {
+					variable = new TemplateVariable(name, type);
+				}
 
 				if (!variable.isRequired() && start < baseUriEndIndex) {
 					baseUriEndIndex = start;
@@ -286,6 +294,9 @@ public class UriTemplate implements Iterable<TemplateVariable>, Serializable {
 		}
 
 		switch (variable.getType()) {
+			case COMPOSITE_PARAM:
+				appendComposite(builder, variable.getName(), value);
+				break;
 			case REQUEST_PARAM:
 			case REQUEST_PARAM_CONTINUED:
 				builder.queryParam(variable.getName(), value);
@@ -297,6 +308,21 @@ public class UriTemplate implements Iterable<TemplateVariable>, Serializable {
 			case FRAGMENT:
 				builder.fragment(value.toString());
 				break;
+		}
+	}
+
+	private static void appendComposite(UriComponentsBuilder builder, String name, Object value) {
+		if (value instanceof Iterable) {
+			for (Object valuePart : (Iterable) value) {
+				builder.queryParam(name, valuePart);
+			}
+		} else if (value instanceof Map) {
+			Set<Map.Entry> valueEntries = ((Map) value).entrySet();
+			for (Map.Entry queryParam : valueEntries) {
+				builder.queryParam(queryParam.getKey().toString(), queryParam.getValue());
+			}
+		} else {
+			builder.queryParam(name, value);
 		}
 	}
 }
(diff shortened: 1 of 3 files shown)


src/main/java/org/springframework/hateoas/UriTemplate.java
```diff
@@ -67,6 +67,7 @@ public UriTemplate(String template) {
 			String[] names = matcher.group(2).split(",");
 
 			for (String name : names) {
+				name = name.endsWith("*")? name.substring(0, name.length() - 1) : name;
```
gregturn: I'm not eager to accept a change in the UriTemplate pattern that also includes a special rule outside the regex to enforce it. I'd also have to read the UriTemplate spec to ensure this fits the spec as well.

To my understanding, UriTemplate is a complex spec with multiple levels, hence getting into it needs to have a well defined reason. That's why I am slow to respond to this one.

pivotal-issuemaster:
@JamesTPF Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=484)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=484) to manually synchronize the status of this Pull Request.

src/main/java/org/springframework/hateoas/UriTemplate.java
JamesTPF: The bit of the pattern that is relevant can be found here:
https://tools.ietf.org/html/rfc6570#section-2.4.2
We have been running with a patched version of the library since August with this fix in it.

pivotal-issuemaster:
@JamesTPF Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=484)!

src/main/java/org/springframework/hateoas/UriTemplate.java
gregturn: This bit of code doesn't implement the spec found in RFC6570. For example, if I try to expand the exploding operator per the spec:

```
	@Test
	public void discoversMultipleRequestParamWithStar() {
		UriTemplate template = new UriTemplate("/foo{&bar,foobar*}");

		assertVariables(template,
			new TemplateVariable("bar", VariableType.REQUEST_PARAM_CONTINUED),
			new TemplateVariable("foobar", VariableType.REQUEST_PARAM_CONTINUED));

		String foo = template.expand(new HashMap<String, Object>(){{
			put("bar", "barExpanded");
			put("foobar", new HashMap<String, String>(){{
				put("city", "Clarksville");
				put("state", "TN");
			}});
		}}).toString();
		System.out.println(foo);
	}
```
...then I don't see what's expected:

```
/foo?bar=barExpanded&foobar=%7Bcity=Clarksville,%20state=TN%7D
```

Per the spec, it should be `/foo?bar=barExpanded?city=Clarksville&state=TN`. Meaning this doesn't appear to be much of a solution.

JamesTPF:
@gregturn 

> Since URI Templates do not contain an indication of type or schema,
   the type for an exploded variable is assumed to be determined by
   context.  For example, the processor might be supplied values in a
   form that differentiates values as strings, lists, or associative
   arrays.  Likewise, the context in which the template is used (script,
   mark-up language, Interface Definition Language, etc.) might define
   rules for associating variable names with types, structures, or
   schema.

So, I've added code that handles
* Lists -> expands to multiple query params (should this handle sets/collections too?)
* Associate Arrays -> maps expand to query params as (key,value) pairs

The last part is a bit confusing: 
> the context in which the template is used (script,
   mark-up language, Interface Definition Language, etc.) might define
   rules for associating variable names with types, structures, or
   schema.

I'm not sure if we need to do anything to cover this?

Also, if the value given for a composite value is not a List and not a Map then the variable does not get expanded.  This seems incorrect.  Should it just add the whole value otherwise?
e.g. on line 324 of https://github.com/spring-projects/spring-hateoas/pull/484/files#diff-36bfaf1489cd680508dc032d1456c320 should it be:
```
else {
   builder.queryParam(name, value);
}

File: src/main/java/org/springframework/hateoas/UriTemplate.java [hunk 2]
```diff
@@ -299,4 +310,17 @@ private static void appendToBuilder(UriComponentsBuilder builder, TemplateVariab
 				break;
 		}
 	}
+
+	private static void appendComposite(UriComponentsBuilder builder, String name, Object value) {
+		if (value instanceof List) {
```
gregturn: I would use `Iterable` instead of `List` since that is the type supported by the Java 5 for loop.

src/test/java/org/springframework/hateoas/UriTemplateUnitTest.java
```diff
@@ -261,6 +261,57 @@ public void allowsAddingTemplateVariable() {
...
+		UriTemplate template = new UriTemplate("/foo{&bar,foobar*}");
+
+		assertVariables(template,
+				new TemplateVariable("bar", VariableType.REQUEST_PARAM_CONTINUED),
+				new TemplateVariable("foobar", VariableType.COMPOSITE_PARAM));
+
+		String expandedTemplate = template.expand(new HashMap<String, Object>(){{
+			put("bar", "barExpanded");
+			put("foobar", new ArrayList<String>(){{
+				add("foo1");
```
gregturn: Or you can do `Arrays.aslist("foo1", "foo2")`.

JamesTPF:
@gregturn I believe I have addressed all of your concerns.  Thanks.

gregturn:
Out this week. Can look next week.

osvaldopina:
Ccghgfzxglm ct

Em quarta-feira, 29 de março de 2017, Greg Turnquist <
notifications@github.com> escreveu:

JamesTPF:
@gregturn Any more thoughts on this?  It would be nice if we could move back to using the main library rather than our patched one.

gregturn:
I'm knee deep working on this branch => https://github.com/spring-projects/spring-hateoas/tree/rebase/affordances

It might take awhile given it adds several new media types and a new API for defining hypermedia.

PR #497, opened 2016-09-27
Support composed annotations in AnnotationMappingDiscoverer.

mp911de:
We now support composed annotations using `@AliasFor` such as `@GetMapping` that are mixed with `@RequestMapping` on type and method level.

```
@RequestMapping("type")
class MyController {

	@GetMapping("method")
	String method() {
	   ...
	}
}
```


olivergierke:
That's merged and polished, thanks!

kevinjom:
Has this change been released?  @olivergierke

PR #500, opened 2016-09-29
Adding id property to Ext to conform to ALPS spec.  Obvious Fix.

Avonsoft:
Also fixing unit tests to include the additional field.


Avonsoft:
I have added obvious fix to the title of this pull request because the addition of the id field should not functionally change the code.  The coinciding tests were modified only to support the new field.  No new functionality was introduced.

olivergierke:
Merged, thanks!

PR #506, opened 2016-10-27
Fix: JacksonSerializationTest is failing.

progosch:
JacksonSerializationTest is failing due to the differences in line endings.


progosch:
Cannot reproduce failing test on CI. It can be reproduced locally though.

Issue #509, opened 2016-11-04
ControllerLinkBuilder does not handle multiple X-Forwarded-* headers values


smilasek:
Hi,

Spring Cloud has released Camden.SR2 this week. It includes following bug fix from Spring Cloud Netflix project: https://github.com/spring-cloud/spring-cloud-netflix/issues/959.
particular commit: https://github.com/spring-cloud/spring-cloud-netflix/commit/a38b7b71ac8be9608ac2530dac41cd6298d696cf

Based on that `X-Forwarded-Host`, `X-Forwarded-Port` and `X-Forwarded-Proto` could have multiple values separated by comma.

`X-Forwarded-Host` is already handled properly by `ControllerLinkBuilder` (https://github.com/spring-projects/spring-hateoas/blob/642b6736764068ffeee38379629db5fc15cf927d/src/main/java/org/springframework/hateoas/mvc/ControllerLinkBuilder.java#L236-L256).

`X-Forwarded-Port` and `X-Forwarded-Proto` are not.

Additionally for multiple values of `X-Forwarded-Port` header application is throwing `NumberFormatException`because of input values like `8080,443` etc.

otrosien:
@dsyer could you point to the RFC specifying the X-Forwarded-Port behaviour you implemented in https://github.com/spring-cloud/spring-cloud-netflix/commit/a38b7b71ac8be9608ac2530dac41cd6298d696cf ?

otrosien:
... found it, it's probably https://tools.ietf.org/html/rfc7239#section-4 -- but in this RFC only "by", "for", "host" and "proto" are defined. The spec allows extensions (#section-5.5), but basically, with regards to "port" it means, that it is implementation-specific. On the other hand, UriComponentsBuilder from spring-framework deals with comma-separated ports, since 4.2.0.RELEASE:
https://github.com/spring-projects/spring-framework/blob/1f6f0dc101fb425bd63786ef669b0626a6591543/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java#L674 / https://github.com/spring-projects/spring-framework/commit/88a14488a194f3795996ba2ac1bf02af639cd7e9#diff-73d13991926d9c9b00031a62c7e39c95

dsyer:
[...]

PR #519, opened 2016-12-07
ControllerLinkBuilder and multiple X-Forwarded-* headers values. (issue #509)

otrosien:
Open for discussion: UriComponentsBuilder does not respect X-Forwarded-Ssl, and I didn't see any JIRA issue adding this support. So for now, it's a question if spring-hateoas should just be aligned with the feature set of spring-framework, or add its own customizations. This would basically revert the PR for issue #112. I would vote for aligning it with what UriComponentsBuilder supports, but I'm not to decide.
That's why I put a junit.Ignore on them until this is clarified.

### DIFF
--- a/src/main/java/org/springframework/hateoas/mvc/ControllerLinkBuilder.java
+++ b/src/main/java/org/springframework/hateoas/mvc/ControllerLinkBuilder.java
@@ -18,6 +18,8 @@ package org.springframework.hateoas.mvc;
 import lombok.RequiredArgsConstructor;
 import lombok.experimental.Delegate;
 
+import static org.springframework.util.StringUtils.hasText;
+
 import java.lang.reflect.Method;
 import java.net.URI;
 import java.util.Map;
@@ -51,6 +53,7 @@ import org.springframework.web.util.UriTemplate;
  * @author Greg Turnquist
  * @author Kevin Conaway
  * @author Andrew Naydyonock
+ * @author Oliver Trosien
  */
 public class ControllerLinkBuilder extends LinkBuilderSupport<ControllerLinkBuilder> {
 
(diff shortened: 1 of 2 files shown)


otrosien:
Would also supersede PR#466

olivergierke:
Is there a way we can use `UriComponentsBuilder` for everything *but* that special handling, i.e. massaging the builder returned by Spring to match the setup `X-Forwarded-Ssl` defines?

otrosien:
I'll give it a try, so you're saying we should care about X-Forwarded-Ssl, even if UriComponentsBuilder doesn't? Isn't it more appropriate to file an issue upstream?

olivergierke:
I am about to file one right away, I just assume a bit of a longer cycle and am not sure it's gonna be introduced in 4.3 in the first place.

otrosien:
Alright, so we'll implement X-Forwarded-Ssl as a work-around for the meantime.

olivergierke:
I did a bit of research and can't really find a place where the intended behavior is really well defined. I guess I'll refrain from asking that to be added in Spring Framework then and rather just keep our current implementation around (i.e. the workaround) as we should be able to move quicker or even remove that support easier than core Spring Framework can.

src/main/java/org/springframework/hateoas/mvc/ControllerLinkBuilder.java
```diff
@@ -253,55 +255,22 @@ public String toString() {
...
-		ServletUriComponentsBuilder builder = ServletUriComponentsBuilder.fromServletMapping(request);
-
-		ForwardedHeader forwarded = ForwardedHeader.of(request.getHeader(ForwardedHeader.NAME));
-		String proto = hasText(forwarded.getProto()) ? forwarded.getProto() : request.getHeader("X-Forwarded-Proto");
 		String forwardedSsl = request.getHeader("X-Forwarded-Ssl");
 
-		if (hasText(proto)) {
-			builder.scheme(proto);
-		} else if (hasText(forwardedSsl) && forwardedSsl.equalsIgnoreCase("on")) {
+		UriComponentsBuilder builder = UriComponentsBuilder.fromHttpRequest(new ServletServerHttpRequest(request));
```
olivergierke: Don't we slightly change the logic here? Previously, if a protocol header was defined in any way, the SSL setting was disregarded. I think we'd override it here, wouldn't we?

otrosien:
@olivergierke I think I'm done. On a side-note: ControllerLinkBuilder is one of five files using windows newlines...

olivergierke:
That's polished and merged. I also added you to the team so that you should be able to push issue branches to the repo directly (i.e.  `/issues/${number}`) and I can actually assign tickets to you. :)

PR #534, opened 2017-01-06
Fixed typo in reference documentation

cortextual:
The documentation describes "one small twak". As far as I can tell, "twak" has a few different meanings, none of which seem appropriate. I think "tweak" was meant.


olivergierke:
That's merged, thanks!

Issue #536, opened 2017-01-31
Remove references to Assert single-arg methods.


christophstrobl:
Single arg ``Assert``s are deprecated for Spring 4.3.7; had been removed temporarily in 5.0 snapshots (are back now), and are object of removal for at least Spring 5.1.

see: [SPR-15196](https://jira.spring.io/browse/SPR-15196)

olivergierke:
That's merged, thanks!

PR #537, opened 2017-01-31
Remove references to Assert single-arg methods.

christophstrobl:
Remove usage of single-arg ``Assert``s by providing message.

Fixes: #536


olivergierke:
That's merged, thanks!

Issue #417, opened 2015-11-20
Support optional Link properties of HAL Spec like "deprecation"


fabianbuch:
HAL specifies additional properties to rel, href and templated in Section 5:

https://tools.ietf.org/html/draft-kelly-json-hal-07#section-5

Is it possible to support these by extending the LinkMixin with "type", "deprecation" and "name"? This would allow deprecating links to deprecated APIs and clients to e. g. automatically log a warning when using deprecated APIs.

wallner:
This is something we would also be very interested in, especially in the optional 'title' attribute. Would you consider a patch, if we came up with one?

tkrille:
+1, we need the `title` attribute too.

olivergierke:
Fixed with #567.

PR #238, opened 2014-08-26
Modify the Link class to support RFC 5988 Target Attributes

jstano:

PR #560, opened 2017-03-22
Add extra attributes to Link and use for HAL mediatype

### LINKED ISSUE #100: Also support optional HAL Link fields
The HAL support so far is great but for it to be really usable the optional link fields defined by the HAL specification should also be available.

There is a pull request #79 about adding a type and title but there are allot more optional fields defined by [HAL](http://tools.ietf.org/html/draft-kelly-json-hal). Not sure if adding all the fields to the Link class is the best solution but that would be a quick fix.

So far I've used a extended version of the Link class that adds the fields (and a fix for the issue with selecting the appropriate JsonSerializer, also noticed by #88). But that is far from ideal and makes allot of the convenience method unusable.

Any thoughts on how this could be implemented in a reasonably unobtrusive manner?

gregturn:
Adds many additional attributes defined in RFC5988 and verifies they work properly in the neutral representation of Link while also being rendered properly in the HAL module.

Additionally, some of the files had inconsistent EOL characters. This patches that as well. To see the differences WITHOUT this bit, add **?w=1** to the github page showing the delta, and whitespace differences will be filtered out.

Resolves #100,#417,#235,#240,#238,#223


src/main/java/org/springframework/hateoas/Link.java
```diff
@@ -1,295 +1,523 @@
...
+	 *
+	 * @param attributes
+	 * @return
+	 */
+	private static Set<String> unrecognizedHeaders(final Map<String, String> attributes) {
+
+		// Copy the existing keys to avoid damaging the original.
+		Set<String> unrecognizedHeaders = new HashSet<String>() {{
+			addAll(attributes.keySet());
+		}};
```
gregturn: Consider using `EnumSet.of(/* items */)` instead.

gregturn:
PR needs to be started from the beginning to avoid gobs of unnecessary changes.

PR #580, opened 2017-05-08
Update IANA relation types

rolandKulcsar:
Updated IANA relation types, see: https://www.iana.org/assignments/link-relations/link-relations.xhtml

PR #584, opened 2017-05-17
#580 - Update IANA relation types

gregturn:
Per https://www.iana.org/assignments/link-relations/link-relations.xhtml, the list of standardized IANA rels has grown.


gregturn:
Resolved via https://github.com/spring-projects/spring-hateoas/commit/483f8479bd1021fb744ba49c11884086784ffe2e

PR #295, opened 2015-02-13
issue #288 Allow links on particular rels to be displayed as an array even if there is only one link

vivin:
This change allows links behind certain rels to always be wrapped by an array regardless of cardinality.

This is related to #291. Instead of trying to control the wrapping based on a resource-rel combination, this approach controls the representation of the link simply based on the rel. Hence, if the specified rels appear in _any_ resource, they will always be wrapped inside an array regardless of cardinality. 

I think this allows for the development of more-consistent representations. If there is ever more than one link behind a particular rel (under any resource), I think it makes more sense to maintain that consistency across all resources by ensuring that the links under that rel are always wrapped in an array.

PR #588, opened 2017-05-22
#288 - Opt to render single links as an array.

gregturn:
Provide the means to render a single link entry as a JSON Array.

Original pull-request: #295
Related issues: #291, #115


gregturn:
Related to #115

olivergierke:
What if we allowed users to rather register a bean `HalConfiguration` or the like — which we'd of course default in our configuration — on which they can configure HAL-specific settings. That link relation and multiplicity issue seems very HAL specific which makes me think it's to specific to justify a top level configuration property in `@EnableHypermediaSupport`. Looks decent beyond that to me.

PR #589, opened 2017-05-23
#470 - Add HAL_JSON_UTF8 to MediaTypes

gregturn:


src/main/java/org/springframework/hateoas/core/JsonPathLinkDiscoverer.java
```diff
@@ -178,6 +178,16 @@ private JsonPath getExpression(String rel) {
 	 */
 	@Override
 	public boolean supports(MediaType delimiter) {
-		return this.mediaType == null ? true : this.mediaType.isCompatibleWith(delimiter);
+
+		if (this.mediaTypes == null) {
```
olivergierke: This by definition will never happen, I think we have to check for an empty collection here.

src/test/java/org/springframework/hateoas/mvc/TypeReferencesIntegrationTest.java
```diff
@@ -82,12 +83,17 @@ public void setUp() {
 	@Test
 	public void usesResourceTypeReference() {
 
-		server.expect(requestTo("/resource")).andRespond(withSuccess(RESOURCE, MediaTypes.HAL_JSON));
```
olivergierke: Can we have separate test cases instead of overloading the current ones?

src/main/java/org/springframework/hateoas/core/JsonPathLinkDiscoverer.java
gregturn: Fixed.

src/test/java/org/springframework/hateoas/mvc/TypeReferencesIntegrationTest.java
gregturn: No problem.

gregturn:
Superceded by https://github.com/spring-projects/spring-hateoas/commit/312134ee823393d5d7fd00fcf54fe7d421cf6310

Issue #572, opened 2017-03-31
Enhance ResourceAssemblerSupport to allow generics


gregturn:
Currently, ResourceAssemblerSupport requires making the domain object either subclass ResourceSupport (like `EmployeeResource extends ResourceSupport`) or creating a separate class (like `class EmployeeResource extends Resource<Employee>`).

If we could extract the core code that presumes a class signature of `ResourceAssemblerSupport<Employee>`, presuming the target **resource** type is `Resource<Employee>`, that would make many things simpler.

gregturn:
Resolved via https://github.com/spring-projects/spring-hateoas/commit/a56392ac7c48b57322c71ef9c2592dcffe2a1354.

Issue #416, opened 2015-11-19
ResourceAssemblerSupport.toResources() should return Resources<D> instead of List<D>


KamikaZeeFu:
When exposing a Collection as a Resource, it is counter intuitive to have to wrap the result of toResources() into an Object of Resources.

```
List<DomainObject> objects = domainRepository.findAll();
List<DomainResource> resources = myResourceAssembler.toResources(objects);
Resources<DomainResource> resourcesWrapped = new Resources<>(resources);
//Add links to resources etc...
```

Ideally, toResources should actually return Resources, like the name implies.

PR #591, opened 2017-05-23
#572 - Create SimpleResourceAssembler for Resource<T>

gregturn:
Implement a ResourceAssembler that returns a Resource<T> solely based on a non-Resource domain class.

Related issues: #416

### DIFF
--- a/src/main/java/org/springframework/hateoas/SimpleResourceAssembler.java
+++ b/src/main/java/org/springframework/hateoas/SimpleResourceAssembler.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.hateoas;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.springframework.util.Assert;
+
+/**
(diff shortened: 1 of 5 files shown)


src/main/java/org/springframework/hateoas/SimpleResourceAssembler.java
```diff
@@ -0,0 +1,84 @@
...
+		addLinks(resource);
+		return resource;
+	}
+
+	/**
+	 * Define links to add to every individual {@link Resource}.
+	 *
+	 * @param resource
+	 */
+	default void addLinks(Resource<T> resource) {
```
schauder: I wonder if it would be better to leave this (and the other `addLinks`) method without a default implementation. 

The current NOOP implementation doesn't seem to be the right thing to do only in very obscure cases.
So anybody implementing this interface will have to provide an implementation, but the presence of a default implementation makes it harder to see that.

src/test/java/org/springframework/hateoas/SimpleResourceAssemblerTest.java
```diff
@@ -0,0 +1,108 @@
...
+import lombok.Data;
+
+import java.util.Collections;
+
+import org.junit.Test;
+
+/**
+ * @author Greg Turnquist
+ */
+public class SimpleResourceAssemblerTest {
```
schauder: Test class naming seems to be a little inconsistent, but I think it should be `...UnitTest`.

Side note, since you are aiming for Spring Data style: that would be `...UnitTests`. Of course, that would need a change in the surefire configuration and should be done in a separate change.

gregturn:
@schauder I'll fix the blank line that emerged in `IdentifiableResourceAssemblerSupport` but delegate the rest of repairs in that test class to #670. There appear to be multiple issues with `IdentifiableResourceAssemblerSupport` that need to be addressed separately.

I think you merely tuned the test cases to expose that.

Regarding renaming stuff as `...UnitTest`, I have opened #777 to track it.

src/main/java/org/springframework/hateoas/SimpleResourceAssembler.java
gregturn: After using it, I agree. So I'll remove the default methods and tune the test cases to match.

gregturn:
Resolved via https://github.com/spring-projects/spring-hateoas/commit/a56392ac7c48b57322c71ef9c2592dcffe2a1354.

Issue #594, opened 2017-06-07
Guard against Spring 5's Not Nullable policy


gregturn:
Spring 5 has instituted guards against accepting nulls. Spring HATEOAS needs to make adjustments since certain values previously returned as `null` are now throwing exceptions.

gregturn:
This is needed to complete https://jira.spring.io/browse/DATAREST-1088

PR #595, opened 2017-06-07
#594 - Protect against Spring 5 non-null inputs

gregturn:
Spring 5 has adopted a "no nulls" policy that causes certain methods to throw exceptions instead of returning a null. This updates Spring HATEOAS to handle it suitably.


olivergierke:
That's merged.

PR #602, opened 2017-06-15
#361 - Add support for resolving values in request mappings

gregturn:


src/main/java/org/springframework/hateoas/mvc/ControllerLinkBuilder.java
```diff
@@ -53,17 +50,21 @@
  * @author Kevin Conaway
  * @author Andrew Naydyonock
  * @author Oliver Trosien
+ * @author Josh Ghiloni
+ * @author Greg Turnquist
  */
 public class ControllerLinkBuilder extends LinkBuilderSupport<ControllerLinkBuilder> {
 
 	private static final String REQUEST_ATTRIBUTES_MISSING = "Could not find current request via RequestContextHolder. Is this being called from a Spring MVC handler?";
-	private static final CachingAnnotationMappingDiscoverer DISCOVERER = new CachingAnnotationMappingDiscoverer(
```
olivergierke: What's the reason we move the caching of the templates into the `LinkBuilder` and not leave it in the `MappingDiscoverer`. I can see that you only want the raw templates to be cached and not necessarily the values evaluated against the `Environment` (as the latter might change), but I still think this could be solved at construction. Maybe a builder that you hand in an annotation type, that'd wrap the `AMD` with the caching one. If you then add an `Environment` to the builder, it could put the property resolving one in front?

gregturn: The issue is all the static fields. The solution was to have a setter method such that it would fetch a default static, unless the setter had overridden a field.

Moving things around, I was able to put the caching back inside CachingANnotationMappingDiscoverer, but at that point, I ended up STILL fetching the final route after property resolution before caching. So I don't see how to do this.

gregturn:
I rebased this PR but haven't resolved the issue you mentioned @olivergierke

gregturn:
@olivergierke I found a solution. Working on updating this PR.

gregturn:
@olivergierke I rewrote this so it doesn't move around the `CachingAnnotationMappingDiscoverer` as before. Instead, `CachingAnnotationMappingDiscoverer` is able to spot if a custom `MappingDiscoverer` has been injected and then use that instead.

gregturn:
I have cleaned up this PR and rebased against latest changes. The idea is to offer a static method through which you can register a Spring `PropertyResolver` so that Spring can resolve variables. Static is the requirement to hook into link creation.

src/main/java/org/springframework/hateoas/mvc/ControllerLinkBuilder.java
```diff
@@ -102,17 +105,26 @@
...
+	ControllerLinkBuilder(UriComponents uriComponents, TemplateVariables variables, MethodInvocation invocation, MappingDiscoverer discoverer) {
 
 		super(uriComponents);
 
 		this.variables = variables;
+		ControllerLinkBuilder.delegateDiscovererOverride = discoverer;
 		this.addAffordances(findAffordances(invocation, uriComponents));
 	}
 
+	public static void setDelegateDiscoverer(MappingDiscoverer delegateDiscovererOverride) {
```
schauder: Some JavaDoc here would be helpful, to point out that this disables caching of `UriTemplate`s

src/main/java/org/springframework/hateoas/mvc/ControllerLinkBuilder.java
```diff
@@ -368,14 +380,72 @@ private static HttpServletRequest getCurrentRequest() {
 	@RequiredArgsConstructor
 	private static class CachingAnnotationMappingDiscoverer implements MappingDiscoverer {
```
schauder: I find the name kind of misleading since the class seems to do two things:

1. determine the correct delegate to use.

2. caching `UriTemplates`

I see various ways to improve that: 2 classes, lengthy name or some JavaDoc.

odrotbohm:
I'm still not convinced we should go down that route. Reconfiguring static properties on calling a setter that is. The main reason is that this is totally unexpected to Spring users, as configuration is always per instance.

I'm in favor of documenting the lack of support for this for static usage of `ControllerLinkBuilder` and rather apply Greg's suggested changes (finding some means to configure a `PropertyResolver`) on `ControllerLinkBuilderFactory` and encourage developers to rather inject that into controllers for non-static usage.

gregturn:
Resolved via https://github.com/spring-projects/spring-hateoas/commit/c5f4bfa0b581139e386df501b7dd959a61ead1e0 and #1328.

PR #581, opened 2017-05-13
Affordances API + HAL-Forms

gregturn:
* Adds new Affordances API
* Introduces HAL-Forms, which uses affordances to automatically generate HTML form data based on marked up domain objects and controllers
Attempted rebase of review/affordances against master

Original pull-request: #340, #447
Related issues: #503, #334

Issue #503, opened 2016-10-14
Spring HATEOAS Forms support


anderruiz:
Hi to all, we’ve been working (Hdiv Security Team) in an extension for Spring HATEOAS project to support FORMs using several different implementations, i.e. XHTML, HAL-Forms, Siren or Uber. The work is based on Dietrich Schulten initial Affordances implementation. Our changes were pushed to Hdiv-Affordances branch in Spring HATEOAS for Oliver Gierke to review them.

A sample app working with Spring HATEOAS and HAL-Forms is available here:

https://github.com/hdiv/hdiv-bank-for-services -> master

We are currently analyzing the best way to integrate it with Spring Data REST by modifying the RESTBucks showcase also

Anyone that could be interested please contact us without hesitation, all suggestions are welcomed :)

gregturn:
Whereas I support the idea of HAL-FORMS, Siren, and Uber, I don't think XHTML is a good idea. That mediatype is quite dated and given this project is mostly JSON oriented, not sure the value investing in this mediatype.

gregturn:
Superceded by #612 and resolved via https://github.com/spring-projects/spring-hateoas/commit/70448a854098343897fd99e3227b4a61db54b921.

Final solution isn't precisely the same thing, but we plan to iterate over it in the near future and expand to support other mediatypes. Feel free to check the HAL-FORMS section of code and their corresponding unit tests.

Also, see monitor [Spring HATEOAS Examples repo](https://github.com/spring-projects/spring-hateoas-examples) as we plan to add an Affordances example very soon.

Issue #334, opened 2015-04-17
Add SIREN support


lucaguada:
Any plan to add support for Siren? or a fluent API in order to generate HATEOAS entities in such way?

it would be interesting to add an integration with siren4j - https://github.com/eserating/siren4j.

nickgrealy:
+1

ilyes:
That would be great

groll75:
+1

dschulten:
Just released [spring-hateoas-ext](https://github.com/dschulten/hydra-java#affordancebuilder-for-rich-hyperlinks) 0.2.0-beta5 to Maven central, it has a [SirenMessageConverter](https://github.com/dschulten/hydra-java#sirenmessageconverter-from-v-020-beta5).

Configuration sample: https://github.com/dschulten/hydra-java/blob/master/hydra-sample/service/src/main/java/de/escalon/hypermedia/sample/Config.java

It is based on AffordanceBuilder, which can also be found at the spring-hateoas [affordances feature branch](https://github.com/spring-projects/spring-hateoas/tree/feature/affordances). I will merge in the latest changes from spring-hateoas-ext into the feature branch during the next days.

andifalk:
+1

sintetico82:
+1

mozinrat:
+1, I hope spring team really looks at this, its being overlooked for more than 2 years.

gregturn:
With Affordances added via #612 (https://github.com/spring-projects/spring-hateoas/commit/70448a854098343897fd99e3227b4a61db54b921), we will be able to move forward with adding support for SIREN.

gregturn:
You may wish to revisit this light of https://github.com/spring-projects/spring-hateoas/issues/833

lazee:
What happened to the Siren support? Seems like all the other issues that blocked this are closed or fixed?

gregturn:
As a project we slowed on implementing new media types and instead shifting to other aspects like affordances, renaming “resource” to “model” and WebFlux support. 

But by making it easy for others to [plug-in their own media type](https://docs.spring.io/spring-hateoas/docs/1.1.0.M3/reference/html/#mediatypes.custom), nothing is stopping someone else from coding and publishing a module that supports SIREN (or whatever).

ingogriebsch:
The following is probably of interest to some of you: https://github.com/ingogriebsch/spring-hateoas-siren

odrotbohm:
[...]

PR #612, opened 2017-07-28
#340 - Adds new Affordances API + HAL-Forms mediatype

gregturn:
* Introduces new Affordances API to build links related to each other to serve other mediatypes
* Introduces HAL-Forms, which uses affordances to automatically generate HTML form data based on Spring MVC annotations.

Original pull-request: #340, #447, #581
Related issues: #503, #334, #71


gregturn:
Also related to #71.

gregturn:
@olivergierke I rebased this PR and got all the test cases to work (again). Also migrated many MORE test cases to AssertJ in the process. It has probably undone some of your polishing from the 2nd commit.

Additionally, I used the IDE to complete several _other_ polishings that are Java 8-based, and that is the 3rd commit. I can understand if you'd prefer cherry picking that separately.

Also, all of this work somehow caused an ALPS test case to break simply due to Jackson and property ordering. I looked up the ALPS spec and adjusted the serializer to order things properly and patched the test case. Again, I can understand opening a separate issue and cherry picking that commit as well.

As a side effect of reading the ALPS spec, it appears the "descriptors" should actually be "descriptor" so I opened #665 but opted to not go THAT far out of scope. :)

Hopefully we are close to merging things, however that needs to be done ultimately. Once done, I can certainly invest some effort in Spring Data REST's Lovelace version to enhance all those controllers to leverage affordances and have an instant increase in usability before circling back to more mediatypes. ;)

Issue #352, opened 2015-05-31
ControllerLinkBuilder.linkTo fails to build a link to a method that uses entity as a target of @PathVariable in its signature


vtsukur:
Consider a project using Spring Data JPA / Spring Data REST to manage an entity `MyEntity` via repository annotated with `@RepositoryRestResource`:

```
@RepositoryRestResource
public interface MyEntityRepository extends PagingAndSortingRepository<MyEntity, Long> {}
```

Consider the following (custom) Spring MVC controller in such an application:

```
@RestController
@RequestMapping("/path/{id}/subpath")
public class MyResourceController {

    @RequestMapping(method = RequestMethod.GET)
    public ... getSomething(@PathVariable("id") MyEntity myEntity) ...

}

```

and the following `ResourceProcessor`:

```
@Component
public class MyResourceProcessor implements ResourceProcessor<Resource<MyEntity>> {

    @Override
    public Resource<MyEntity> process(Resource<MyEntity> resource) {
        resource.add(linkTo(methodOn(MyResourceController.class).getSomething(resource.getContent())).withRel("payment"));
        return resource;
    }

}
```

`ResourceProcessor` will fail to generate the link avoiding the knowledge and capabilities of `DomainClassConverter` and `@PathVariable`

```
org.springframework.core.convert.ConverterNotFoundException: No converter found capable of converting from type @org.springframework.web.bind.annotation.PathVariable org...domain.MyEntity to type java.lang.String
        at org.springframework.core.convert.support.GenericConversionService.handleConverterNotFound(GenericConversionService.java:313)
        at org.springframework.core.convert.support.GenericConversionService.convert(GenericConversionService.java:195)
        at org.springframework.hateoas.mvc.AnnotatedParametersParameterAccessor$BoundMethodParameter.asString(AnnotatedParametersParameterAccessor.java:172)
        at org.springframework.hateoas.mvc.ControllerLinkBuilderFactory.linkTo(ControllerLinkBuilderFactory.java:130)
        at org.springframework.hateoas.mvc.ControllerLinkBuilder.linkTo(ControllerLinkBuilder.java:136)
        at org...MyResourceProcessor.process(BookingProcessor.java:25)
        at org...MyResourceProcessor.process(BookingProcessor.java:16)
```

gregturn:
For a Spring data rest project, you should not use RedtController. See https://github.com/spring-projects/spring-data-rest/pull/177 for documentation updates on the proper way to build overriding controllers.

[...]

Issue #149, opened 2014-02-07
AnnotatedParametersParameterAccessor uses DefaultFormattingConversionService


nickspacek:
Unless I'm missing something, the use of DefaultFormattingConversionService means there is no way to support custom parameters in MVC Controllers using (for example) linkTo/methodOn. If I have a controller method like this:

```
@RequestMapping(/{vegetableId}
public void get(@PathVariable("vegetableId") Vegetable vegetable)
```

Then even if I register a custom Converter to convert the object to a string representation, the linkTo/methodOn won't work because the Converter isn't part of the DefaultFormattingConversionService. Shouldn't the AnnotatedParametersParameterAccessor be referencing a ConversionService from the ApplicationContext?

odrotbohm:
Duplicates #118.

Issue #118, opened 2013-11-11
BoundMethodParameter references static, default instance of ConversionService


nebhale:
Currently the BoundMethodParameter class references a [static, default instance](https://github.com/spring-projects/spring-hateoas/blob/master/src/main/java/org/springframework/hateoas/mvc/AnnotatedParametersParameterAccessor.java#L81) of `ConversionService`.  This means that should a user add `Converter`s to their `ApplicationContext` they will not be picked up and used.  This should be changed such that a `ConversionService` specified in the `ApplicationContext` will be used, otherwise the default will be.

mhuffman-r7:
+1

bostond:
+1

hobo05:
+1

ismailbay:
+1

tufelix:
+1

francesc79:
+1

osvaldopina:
I developed a [framework](https://github.com/osvaldopina/linkbuilder) to render links in spring hateoas projects. It also supports rendering links for user defined types. An explanation can be found in this [stackoverflow answer](http://stackoverflow.com/a/37714119/5144732)

snowe2010:
Is there an end in sight for this bug? It's been open for almost 2 years and none of the provided solutions have worked in our situation.

bobonaut:
+1

aksdb:
+1

galovics:
Is there any progress here?

raulgomis:
@olivergierke any progress here?

lcaille:
+1

rgacki:
+1

This hit me hard at the end of a project. This needs to be fixed!!!

gregturn:
Need to figure out a pared back version of this. The issue is against Spring Data JPA + Spring Data REST, so it's current incarnation requires hooking together lots of components. If there is a Spring HATEOAS-specific issue, need a much simpler case to illustrate.

rgacki:
A simple case:

I have controller method parameter with custom types that are annotated with annotations such as @PathVariable or @RequestParam. While Spring MVC is behaving correctly by using my custom `org.springframework.core.convert.converter.Converter`s, spring-hateoas is not using these converters when Links to the methods are built with `org.springframework.hateoas.mvc.ControllerLinkBuilder` and values are passed to the proxied method call.

A (simplified) code example:

```
@Controller
@RequestMapping("/users/{userId}")
public class UserProfileController {

[...]

PR #618, opened 2017-08-28
#118 - Allow injecting custom ConversionService

gregturn:
Existing ConversionService is fixed as a static inside BoundMethodParameter. This introduces ability to inject an alternative ConversionService.

Related issues: #352, #149


gregturn:
To utilize this in a Spring MVC application context driven fashion, we'd probably need a `BeanPostProcessor` that scans for conversion services, and then makes the static call to inject it using the method provided in this PR.

gregturn:
@grtessman I'm actually investigating a new approach. Based on the efforts on #728, I am looking at more comprehensive way to define your own hypermedia handler and registering it for both Spring MVC and Spring WebFlux. If successful, we won't need this approach at all.

Issue #617, opened 2017-08-17
Parameter names not detected in case @AliasFor'ed attributes are used


caoyuanqi:
the code like below, 

```
@RequestMapping(value = "/account/{accountName}",method =RequestMethod.GET)
    public AccountResource getAccountInfo(@P("u") @PathVariable String accountName) {
        AccountResource resource = new AccountResource();
        Account account = accountService.getAccountInfo(accountName);
        resource.setAccountName(account.getAccountName());
        resource.setType(account.getType());
        String investorId = account.getInvestor().getId();
        log.info("investor id is " + investorId);
        Link selfLink = linkTo(methodOn(AccountController.class)
                .getAccountInfo(accountName)).withSelfRel();
        Link investorLink = linkTo(methodOn(InvestorController.class).getInverstor(investorId)).withRel("investor");
       // Link investorLink = new Link(baseUrl + serviceUrl
        //        + "/investor/" + investorId).withRel("investor");
        resource.add(selfLink);
        resource.add(investorLink);
        if (accountService.getInvestmentProfile(accountName) != null){
            Link investmentProfileLink = new Link(baseUrl + serviceUrl
                    + "/" + accountName + "/profile").withRel("mandate");
            resource.add(investmentProfileLink);
        }
        return resource;
    } 
```
Return Json is 

``` 
{
  "type": "investor",
  "accountName": "cc1502762479261",
  "_links": {
    "self": {
      "href": "http://localhost:2222/accounts/account/ccxxxxxxxxx"
    },
    "investor": {
      "href": "http://localhost:2222/accounts/investor/{id}",
      "templated": true
    },
    "mandate": {
      "href": "http://localhost:9191/accounts/ccxxxxxxxxxx/profile"
    }
  }
}
``` 
Any ideas?

gregturn:
Show me InvestorController and a log output with the investorId.

caoyuanqi:
Basically, we are coding with Spring cloud. Related code as below:

```
package au.com.othera.accountservice.web;

import java.security.Principal;
import java.util.List;
import java.util.Map;

import javax.validation.Valid;

import com.fasterxml.jackson.databind.node.ObjectNode;

import io.swagger.annotations.ApiOperation;
import io.swagger.annotations.ApiParam;
import io.swagger.annotations.ApiResponse;
import io.swagger.annotations.ApiResponses;
import lombok.extern.slf4j.Slf4j;
import org.springframework.beans.factory.annotation.Autowired;
[...]

PR #621, opened 2017-09-01
#617 - Use SynthesizingMethodParameter to support AliasFor attributes

gregturn:
When parsing the controller method parameters, use `SynthesizingMethodParameter` in order to leverage Spring's `@AliasFor` annotations that bridge attributes together. That way, fetching a parameter's annotation supports aliased attributes.


gregturn:
@olivergierke Rewrote previous solution to instead use `SynthesizingMethodParameter`.

PR #622, opened 2017-09-07
Use "else if" in HypermediaSupportBeanDefinitionRegistrar.postProcessBeforeInitialization()

izeye:
This PR uses `else if` because the conditions are mutually exclusive.


olivergierke:
We try to avoid `else` expressions as they make adding and removing new criteria slightly harder. As there's no semantic difference between the two versions, I'd like to drop that PR.

izeye:
@olivergierke Okay. Thanks for clarification!

Issue #532, opened 2016-12-22
should honor use-hal-as-default-json-media-type


rainoko:
Spring boot introduced configuration use-hal-as-default-json-media-type which helps fallback from hal format links. spring-hateoas should honor this otherwise it is not compatible with boot anymore.

olivergierke:
There's nothing in Spring HATEOAS that it could honor here as it doesn't know anything about Spring Boot. When you say "anymore" can you please clarify what you mean by that? Has something worked before that now doesn't anymore? Would you mind getting a bit more precise on what works and what doesn't? Maybe through a test case/project?

rainoko:
I tried to migrate from 0.18 to 0.23 (0.22 had bug with context)
problem is that 0.23 registretes HAL support for jackson converter.

In general spring-boot autoconfiguration doing same thing but honors flag use-hal-as-default-json-media-type. This registration adds application/json as supported media type for jackson configuration and causing HAL format by default.
org.springframework.boot.autoconfigure.hateoas.HypermediaHttpMessageConverterConfiguration

Now it seems that spring-hateoas doing it also in 
org.springframework.hateoas.config.HypermediaSupportBeanDefinitionRegistrar
But seems that I cant drive it or I'm not clever enough to drive it.

rainoko:
any news about this?

rainoko:
seems that issue not appears anymore.

abobwhite:
@rainoko I am still seeing this issue. I am having a problem with getting `links` on GETs but `_links` on POST and PUT responses. I want it to always be one or the other.

gregturn:
Can you please list:

* Version of Spring Boot
* Version of Spring HATEOAS
* Are you using Spring Data REST (sometimes people use SDR, but open their issues here as related project).
* Any custom configurations you are doing involving message converters, `@EnableHypermediaSupport`
* Do you have a Spring MVC controller and some linkTo(methodOn() type code you are using that is yielding awkward stuff?
* What are the two operations you are performing and where (RestTemplate/curl) that yield different JSON formats?

We need this level of detail to pinpoint what is happening.

abobwhite:
Hi @gregturn,

Thanks for the response. I am trying to use `use-hal-as-default-json-media-type` to make sure I get `links` consistently and not `links` for GET requests and `_links` for POST and PUT requests. I'd be happy if it was either as long as it's consistent.

- Spring Boot  1.5.7.RELEASE
[...]

PR #629, opened 2017-09-18
Experiment: Detect link methods that do NOT return resources

gregturn:
**>>> DO NOT MERGE <<<**

Detect endpoints that do NOT return classes assignable to `ResourceSupport`.

Question: Do we post a warning, or actually break?

### DIFF
--- a/src/main/java/org/springframework/hateoas/mvc/ControllerLinkBuilderFactory.java
+++ b/src/main/java/org/springframework/hateoas/mvc/ControllerLinkBuilderFactory.java
@@ -30,9 +30,12 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.core.MethodParameter;
 import org.springframework.hateoas.Link;
 import org.springframework.hateoas.MethodLinkBuilderFactory;
+import org.springframework.hateoas.ResourceSupport;
 import org.springframework.hateoas.TemplateVariable;
 import org.springframework.hateoas.TemplateVariables;
 import org.springframework.hateoas.core.AnnotationAttribute;
@@ -44,6 +47,7 @@ import org.springframework.hateoas.core.MappingDiscoverer;
 import org.springframework.hateoas.core.MethodParameters;
 import org.springframework.hateoas.mvc.AnnotatedParametersParameterAccessor.BoundMethodParameter;
 import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
 import org.springframework.util.MultiValueMap;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestMapping;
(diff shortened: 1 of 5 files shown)


gregturn:
@olivergierke I wrote some code that detects if the Spring MVC endpoints return type is NOT a `ResourceSupport`/`Resource`/`Resources` type, because it will probably yield the _wrong_ JSON. I've seen a couple issues recently that track back to improperly built endpoints.

In this coding, it logs a warning AND throws a `RuntimeException`. I imagine the warning is good, but perhaps we need to drop the exception? Or perhaps we need an option to toggle the exception?

This looks like a keen mechanism to warn users that they've misconstructed their controllers.

olivergierke:
> @olivergierke I wrote some code that detects if the Spring MVC endpoints return type is NOT a ResourceSupport/Resource/Resources type, because it will probably yield the wrong JSON.

"Wrong JSON"? What problem are we actually solving here? When creating links, we must not care whether the resource we point to is hypermedia enabled or not? How would I point to a controller method serving a PDF? Also, note that it's very common to use `HttpEntity<?>` or `ResponseEntity<?>` as wrapper for the actual payloads.

gregturn:
The problem is that some users of Spring HATEOAS have controllers returning `List<Resource<Item>>`, which the HAL serializers won't render as valid JSON. These people should return `Resources<Resource<Item>>` instead (or some other type that is either ResourceSupport, Resource, or Resources).

By catching these scenarios, we'll stop getting issues opened like #532, that turned out to simply be a bad return type for Spring HATEOAS.

gregturn:
I think part of this is due to `ResourceAssemblerSupport` having `toResources` returning a `List` (an issue patched by #591), and people using that as their controller return type.

Seems like a simple thing we can detect and either warn or fail fast for the user.

olivergierke:
Hm, I see. Maybe I am just not getting why this has anything to do with *link creation*? These are completely orthogonal issues, right? Wrapping an `HttpEntity` into a `Resource` is completely backwards as if at all, the `Resource` is the payload of the `HttpEntity`. But then again, there's no reason we should remove the ability to point to arbitrary controller methods.

The problem of `List<Resource<Item>>` still persists even with that change. If you don't point to that controller method it will still be used and render the "invalid" JSON. TBCH, that's not even invalid, because it might be a deliberate decision to simply render an array of `Resource` instances so that the individual elements carry hypermedia elements, but the wrapping array does not. So there are very valid reasons to use both variants. Maybe it's just a matter of documenting this aspect slightly better?

gregturn:
The reason I picked link creation as the point of detection is because it appeared the only place to sniff out the return type of a Spring MVC controller. 

Well perhaps my check isn't sophisticated enough. If it's a Spring MVC response container, check its payload type?

And your point about being intentionally done by the end user is why I speculated this might best be a logged warning, not an exception, perhaps with an option to disable?

When I say option to disable, perhaps an annotation that signals "I know what I'm doing" on the endpoint?  @SupressResourceValidation or something?

gregturn:
Closed in light of #416.

PR #653, opened 2017-10-15
Use Java 8 Stream API

rolandKulcsar:
Refactor to use Java 8 Stream API and the diamond operator in the modified files.


olivergierke:
That's merged as well. Thanks!

Issue #665, opened 2017-11-06
ALPS spec name is "descriptor" not "descriptors"


gregturn:
ALPS serialization generates "descriptors", when the value in the spec is singular.

https://tools.ietf.org/html/draft-amundsen-richardson-foster-alps-01#section-2.2.3

PR #666, opened 2017-11-07
#665 - Fix ALPS format

gregturn:
According to the ALPS spec, it's "descriptor" not "descriptors", despite it depicting an array of potentially multiple values. Also fixed the ordering to match as closely as possible all the samples shown in the spec.


olivergierke:
That's merged.

Issue #667, opened 2017-11-14
Templated URIs in a HAL document cannot be deserialized


gregturn:
Attempting to deserialize a HAL document containing the following structure will fail due to the "templated" property not being a real property of `Link`.

```
{
   "_links" : {
     "order" : {
       "href" : "/orders{?id}",
       "templated" : true
     }
   }
 } 
```

gregturn:
Resolved via https://github.com/spring-projects/spring-hateoas/commit/d5649c0de16f3067c1383ce1a373bd6d311f4440.

odrotbohm:
Isn't that a question of setting the Jackson `ObjectMapper` to ignore unknown properties?

gregturn:
I’ll test if putting that Jackson annotation on Link works. I vaguely remember trying this and somehow it didn’t work, but maybe that was another Jackson setting I was fiddling with. 

We have a test case that should work properly.

PR #668, opened 2017-11-14
#667 - Fix Link to ignore setting "templated" properties

gregturn:
Deserializing a HAL document with a templated URI will cause Jackson to break due to there being no "templated" property. This PR fixes that by adding a private, no-op `setTemplated` method.


gregturn:
@schauder If you want to look at this one, it's pretty simple.

gregturn:
Resolved via https://github.com/spring-projects/spring-hateoas/commit/d5649c0de16f3067c1383ce1a373bd6d311f4440.

PR #670, opened 2017-11-15
Handle Optional in IdentifiableResourceAssemblerSupport.unwrapIdentifiables()

izeye:
This PR also fixes tests having a wrong assertion.

### DIFF
--- a/src/main/java/org/springframework/hateoas/mvc/IdentifiableResourceAssemblerSupport.java
+++ b/src/main/java/org/springframework/hateoas/mvc/IdentifiableResourceAssemblerSupport.java
@@ -68,7 +68,7 @@ public abstract class IdentifiableResourceAssemblerSupport<T extends Identifiabl
 		Assert.notNull(id, "Id must not be null!");
 
 		D instance = instantiateResource(entity);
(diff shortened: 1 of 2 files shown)


gregturn:
I'm not convinced this patches the whole issue. Actually, it's not this PR I'm concerned with but rather the basis for `IdentifiableResourceAssemblerSupport`. See below.

If the controller is:

```
@RequestMapping("/people")
static class PersonController {

}
```

...then a `Person(1L)` will resolve to `/people/1`. That makes perfect sense.

However, for a controller like this:

```
@RequestMapping("/people/{id}/{foo}/addresses")
static class ParameterizedController {

}
```
...the URI generated for `Person(1L)` with an extra param of `bar` yields `/people/1/bar/addresses/1`. I'm pretty sure that the intention of the controller is NOT to have the id value listed twice. But that is how `IdentifiableResourceAssemblerSupport` operates.

It seems like if the `id` has already been used explicitly, it shouldn't be blatantly slapped onto the end _again_.

gregturn:
What are your thoughts on this @olivergierke ?

gregturn:
Thanks @izeye, but we actually removed `Identifiable` back in #842. So this change is no longer needed.

Issue #678, opened 2017-12-06
Add profile member


DavidBiesack:
[draft-kelly-json-hal-08 JSON HAL](https://tools.ietf.org/html/draft-kelly-json-hal-08#section-5.6) defines the optional string `profile` member which is not implemented in `org.springframework.hateoas.Link`.  The profile field was requested in #656 but not implemented in 0.24.RELEASE. Please add it.

(As an aside, advice on how to extend the `org.springframework.hateoas.Link` class would be welcome. Lombok makes this pretty hard, since the `@AllConstructor` is package protected and not callable from the subclass. Without access to theat `super(...)` constructor, I don't see a way to extend `Link` and add the field - there are no setters, only `@Wither` members which do not clone the subclass object but call a `org.springframework.hateoas.Link1` constructor.  Perhaps another enhancement would be to remove `AccessLevel.PACKAGE` )

gregturn:
@DavidBiesack We don't wish to open up the visibility, so instead I've made an update to Link that I'll wait for @olivergierke to review and approve.

DavidBiesack:
thanks for the PR to add `profile`. Since you have elected to not widen the scope of the constructor, is there some other way that one can subclass Link to add new fields?

gregturn:
Link isn't reallly designed as an extension point to the project. In short you could fork and work from that.

PR #679, opened 2017-12-07
#678 - Add 'profile' to Link

gregturn:
Fully support HAL spec by adding "profile" as another optional attribute to `Link`.


gregturn:
@schauder You may additionally wish to start the HAL spec and the RFC 5988 spec (both links provided in the commit message itself).

PR #686, opened 2018-01-03
Support whitespace in Links Header

cbuehler:
The RFC uses [Examples](https://tools.ietf.org/html/rfc5988#section-5.5)
which cannot be parsed correctly


gregturn:
Resolved via https://github.com/spring-projects/spring-hateoas/commit/21719bb0417c9347a12036c0072a73646d396bd3.

Thanks @cbuehler!

PR #695, opened 2018-02-01
Remove Jaxb annotations

gregturn:


gregturn:
@schauder This PR pulls out XML support. Given our HAL format and others never did support XML, and having XML stuff actually confused Spring MVC routes, it makes sense to pull this stuff.

src/test/java/org/springframework/hateoas/hal/RenderHypermediaForDefaultAcceptHeadersTest.java
```diff
@@ -0,0 +1,216 @@
...
+
+	@Before
+	public void setUp() {
+		this.mockMvc = webAppContextSetup(this.context).build();
+	}
+
+	/**
+	 * Verify the default {@literal Accept} header used by a browser (which favors XML/HTML) still yields HAL-JSON.
+	 */
+	@Test
```
schauder: Include a comment referencing an issue in the test or don't you do that for this project?
Applies to all `@Test` annotations.
Also applies to many other classes not touched in this commit.
Maybe cleaning this up an moving to the style of Spring Data is in place?

gregturn: Indeed, Spring HATEOAS observes the Spring Data standards, so using the notation is a good point.

Issue #559, opened 2017-03-21
ControllerLinkBuilder link expansion leads to multiple url-encoding of parameters


rainerfrey:
When building a link with `methodOn`, and supplying values for some parameters but not all, a link template with variables for the remaining parameters is created, while the supplied values are already expanded. When expanding the resulting template, the initially supplied values are url-encoded again, leading to invalid URLs. This affects explicitly calling `expand()` of course, but also the pagination links of the `PagedResourcesAssembler` from Spring Data. Here is some code example:

```java
    @GetMapping("/")
    public Resource test(
            @RequestParam(name = "date1", required = false) @DateTimeFormat(iso = DATE_TIME) Date date1,
            @RequestParam(name = "date2", required = false) @DateTimeFormat(iso = DATE_TIME) Date date2) {
        Link template = linkTo(methodOn(TestController.class).test(date1, null)).withRel("template");
        Link expanded = linkTo(methodOn(TestController.class).test(date1, null)).withRel("expanded").expand();
        return new Resource("", template, expanded);
    }
```
This leads to following response:
```json
{
    "_links": {
        "expanded": {
            "href": "http://localhost:8080/?date1=2017-03-21T18:54:00.000%25252B0000"
        },
        "template": {
            "href": "http://localhost:8080/?date1=2017-03-21T18:54:00.000%2B0000{&date2}",
            "templated": true
        }
    },
    "content": ""
}
```
As you see, the `+` in the date has been encoded correctly in the template, but the `%`-sign has been encoded again in the expanded link, even twice as shown by the additional `25`.

I have a runnable example with a test case in [this repository](https://github.com/rainerfrey/link-encoding-problem)

ripperr:
Same issue here.

ndzeller:
Experiencing problems with this issue as well.

tmbell:
Approaching a year on this bug.  Anybody looking into it?  or found a viable workaround?

jeffreycrump:
This bug makes the ControllerLinkBuilder pretty much useless for creating non-templated links.  Any idea on the progress of this PR?

Issue #593, opened 2017-05-26
UriTemplate handles encoded variables improperly on expand call


keaplogik:
When testing the `UriTemplate`, we ran into several issues with encoding. The URI template, when given a path like `/foo%20bar` will return `/foo%252520bar` on a call to `expand`. It looks like both the hateoas UriTemplate and the spring UriTemplate are not handling ecoded URLs. Also if you provide a URL with a character needing encoding, you get the `%` sign escaped only once.

Here is a spock test to shows the errors. The last three tests will fail due to the issues.

```groovy
@Unroll("template expand with params '#params' and uri '#uri' results in '#expectedUri ")
    def "template expands with parameter map"() {
        when:
        UriTemplate template = new UriTemplate(uri)
        String expandedUri = template.expand(params)

        then:
        expandedUri == expectedUri

        where:
        uri                             | params                      | expectedUri
        '/foo/bar{?x}'                  | ['x': 1]                    | '/foo/bar?x=1'
        '/foo/bar{?x,y}'                | ['x': 1, 'y': "2"]          | '/foo/bar?x=1&y=2'
        '/foo/bar{?x}{&y}'              | ['x': 1, 'y': "2"]          | '/foo/bar?x=1&y=2'
        '/foo/bar?x=1{&y}'              | ['y': 2]                    | '/foo/bar?x=1&y=2'
        '/foo/bar?x=1{&y,z}'            | ['y': 2, 'z': 3L]           | '/foo/bar?x=1&y=2&z=3'
        '/foo{/x}'                      | ['x': 1]                    | '/foo/1'
        '/foo{/x,y}'                    | ['x': 1, 'y': "2"]          | '/foo/1/2'
        '/foo{/x}{/y}'                  | ['x': 1, 'y': "2"]          | '/foo/1/2'
        '/foo{/x}{/y}{?z}'              | ['x': 1, 'y': "2", 'z': 3L] | '/foo/1/2?z=3'
        '/foo/{x}'                      | ['x': 1]                    | '/foo/1'
        '/foo/{x}/bar'                  | ['x': 1]                    | '/foo/1/bar'
        '/services/foo/{x}/bar/{y}/gaz' | ['x': 1, 'y': "2"]          | '/services/foo/1/bar/2/gaz'
        '/foo/{x}/bar/{y}/bar{?z}'      | ['x': 1, 'y': "2", 'z': 3L] | '/foo/1/bar/2/bar?z=3'
[...]

PR #738, opened 2018-09-17
559: fix multiple encodings of params

Climax85:
In order to #559 I got multiple encodings using the ControllerLinkBuilder. I added corresponding tests and fixed the issue. The ControllerLinkBuilder#bindRequestParameters method already encodes the params in the template url. Any further expand of the url will encode the % sign in %25. So no additional encoding should be done.
The Link#expand method is not only used to expand parameters in the template but also to check for missing required parameters.
I added another checkParams method to the UriTemplate class to only check the params but not encoding them.

### DIFF
--- a/src/main/java/org/springframework/hateoas/UriTemplate.java
+++ b/src/main/java/org/springframework/hateoas/UriTemplate.java
@@ -204,22 +204,16 @@ public class UriTemplate implements Iterable<TemplateVariable>, Serializable {
 	}
 
 	/**
-	 * Expands the {@link UriTemplate} using the given parameters. The values will be applied in the order of the
-	 * variables discovered.
+	 * Checks the {@link UriTemplate} for unset but required params.
 	 *
-	 * @param parameters
 	 * @return
-	 * @see #expand(Map)
 	 */
-	public URI checkParams(Object... parameters) {
-		org.springframework.web.util.UriTemplate baseTemplate = new org.springframework.web.util.UriTemplate(baseUri);
-		UriComponentsBuilder builder = UriComponentsBuilder.fromUri(baseTemplate.expand(parameters));
-		Iterator<Object> iterator = Arrays.asList(parameters).iterator();
-
+	public URI checkRequiredParams() {
 		for (TemplateVariable variable : getOptionalVariables()) {
-
-			Object value = iterator.hasNext() ? iterator.next() : null;
-			appendToBuilder(builder, variable, value);
+			if (variable.isRequired()) {
+				throw new IllegalArgumentException(
+					String.format("Template variable %s is required but no value was given!", variable.getName()));
+			}
 		}
 
 		return URI.create(baseUri);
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@Climax85 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=738)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=738) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@Climax85 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=738)!

src/main/java/org/springframework/hateoas/UriTemplate.java
```diff
@@ -203,6 +203,28 @@ public URI expand(Object... parameters) {
...
 
+	/**
+	 * Expands the {@link UriTemplate} using the given parameters. The values will be applied in the order of the
+	 * variables discovered.
+	 *
+	 * @param parameters
+	 * @return
+	 * @see #expand(Map)
+	 */
+	public URI checkParams(Object... parameters) {
```
olivergierke: I am not sure I get the purpose of this method. It'll always return `URI.create(baseUri)`. So what is all the code before supposed to do? The method name is quite misleading to. There's no check going on anywhere. It just returns the base URI. If that alone is worth doing, it could very well be added to `expand(…)` directly.

odrotbohm (changes requested):
I am not sure about this one. The test cases don't seem to check for any double encoding in the first place. Their names aren't quite telling what the intend of the tests are actually.

Also, see my comment on `checkParams(…)`. I didn't actually get the intention of the method.

gregturn:
I second @olivergierke 's comments.

I'd start with a battery of tests exposing the issue at hand. In fact, the code you from #559 would be a perfect candidate to put into a unit test that you could then test away at ensuring it acts properly.

src/main/java/org/springframework/hateoas/UriTemplate.java
Climax85: I need to add the info that I am using the 0.25.0 version of haetoas and as so a 4.3 version of spring-web. In master of spring-web I can see that the encoding has changed and a enum for the encoding state has been added. Maybe this fixes the problem in spring 5.

jeffreycrump:
Any update here?  This bug makes ControllerLinkBuilder pretty much useless for non-template URLs.

gregturn:
@Climax85 @jeffreycrump 

Do you have a test case that exposes the problems here? Also, there have been considerable changes to Spring HATEOAS since this ticket was opened including:

* Migration to `WebMvcLinkBuilder`
* Deprecation of `ControllerLinkBuilder`
* Introduction of `WebFluxLinkBuilder`
* #593 

Please tell me if the latest version (1.1.0.M3 just came out) solves your problems. If not, can you provide a test case illustrating the issue.

Issue #743, opened 2018-10-19
Allow specifying a custom header value for each hop in Traverson


manishpmisra:
Hi All,

I am trying to implement a solution where I want to hop at diffrent level with specific header to each hop. I saw the existing code of Traverson code but it looks like at each hop it is using the same header. Can we add something so we can pass Specific header info to individual hops?

manishpmisra:
Hi @gregturn and @odrotbohm,
I have already resolved this issue in my local and have added a 'withHeader' method in Hop class to provide custom header(s) for each hop. And it is working well. If you guys are okay I would like to commit my code here. Let me know if I can do so and If yes, can you please create a pull request so I can contribute the same.

Thanks

odrotbohm:
Feel free to commit to your fork of the repository and create a pull request 🙃.

manishpmisra:
Sure I will do that. Thanks

manishpmisra:
@gregturn I have created a PR https://github.com/spring-projects/spring-hateoas/pull/747 for the same. My solution is little diffrent and I used the same way we are adding parameters to the Hop. Please take a look.

gregturn:
Resolved via https://github.com/spring-projects/spring-hateoas/commit/b60799d7fa6072282744170293b7f9f4fa793a4f and https://github.com/spring-projects/spring-hateoas/commit/df618efd12006efe0c9ca17413a64b8f0a5b33f1.

PR #744, opened 2018-10-23
Introduce option to apply custom headers on every hop.

gregturn:
* Introduce HeaderUtils to make it easier to build custom set of headers

Related: #743


gregturn:
Resolved via https://github.com/spring-projects/spring-hateoas/commit/b60799d7fa6072282744170293b7f9f4fa793a4f and https://github.com/spring-projects/spring-hateoas/commit/df618efd12006efe0c9ca17413a64b8f0a5b33f1

PR #749, opened 2018-11-12
Performance optimization by extra caching

mst-avaleo:
The PR contains caching for baseUrl, UriTemplate and Mapping. 

Here's results from a simple performance test, which invokes 1000 times a controller which is returning 1000 entities with one self link.
no links - 552ms  
0.25.0 - 57406ms
with PR - 18813ms

### DIFF
--- a/src/main/java/org/springframework/hateoas/core/CachingMappingDiscoverer.java
+++ b/src/main/java/org/springframework/hateoas/core/CachingMappingDiscoverer.java
@@ -1,0 +1,73 @@
+package org.springframework.hateoas.core;
+
+import java.lang.reflect.Method;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+public class CachingMappingDiscoverer implements MappingDiscoverer {
+	private Map<String, String> mappingCache = new ConcurrentHashMap<String, String>();
+	private MappingDiscoverer discoverer;
+
+	public CachingMappingDiscoverer(MappingDiscoverer discoverer) {
+		this.discoverer = discoverer;
+	}
+
+	@Override
+	public String getMapping(final Class<?> type) {
+		String key = key(type, null);
+		return getMapping(key, new CachedCall() {
+			@Override
+			public String getMapping() {
+				return discoverer.getMapping(type);
+			}
+		});
+	}
+
+	@Override
+	public String getMapping(final Method method) {
+		String key = key(method.getDeclaringClass(), method);
+		return getMapping(key, new CachedCall() {
+			@Override
+			public String getMapping() {
+				return discoverer.getMapping(method);
+			}
+		});
+	}
+
+	@Override
+	public String getMapping(final Class<?> type, final Method method) {
+		String key = key(type, method);
+		return getMapping(key, new CachedCall() {
+			@Override
+			public String getMapping() {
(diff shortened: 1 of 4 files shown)


pivotal-issuemaster:
@mst-avaleo Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=749)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=749) to manually synchronize the status of this Pull Request.

gregturn:
@mst-avaleo I'm really excited for your PR. Can you sign the CLA?

gregturn:
@mst-avaleo The CLA tool may have been broken. If possible, can you try to sign the CLA so we can proceed with processing your PR?

mst-avaleo:
We are in the middle of signing the CLA on corporate level. When that's done I want to post another PR, which has more invasive changes but with bigger performance improvement.

gregturn:
Great! 

BTW, I’d suggest doing future work against the master branch. That version is based on Java 8 and has the most freedom for change. We are also limiting the scope of backports to 0.25.x and 0.x.

mst-avaleo:
BTW. I've did tests against version master branch, and it's twice slower than version 0.25.0. I haven't spent more time on that, since 0.25.0 is still GA version.

gregturn:
@mst-avaleo Given we are waiting for a corporate signing of our CLA, would you like to use this time to dig a little deeper into why this doesn't work as well in 1.0.0.BUILD-SNAPSHOT, given this is the version I'd rather fold your change into?

odrotbohm:
Coming back to this I have a couple of questions:

1. Can you share the benchmarks you have to verify the effect of the changes? Can you share what percentage of the speedup is related to the cached mappings VS. the one caching the base URI VS. caching the `UriTemplate`s?
2. What's the reason you're not using `computeIfAbsent(…)` in the implementation of `CachingMappingDiscoverer`?

odrotbohm:
To answer 2. myself: because the PR is tailored towards 0.x and that's still on Java 6.

odrotbohm:
I've polished and forward ported this to the master branch for inclusion in 1.0 as well.

mst-avaleo:
@gregturn @odrotbohm Sorry for not answering. I wanted to wait until the CLA will be signed first.

Thanks for including the changes in 0.25.x. I've checked them with my simple benchmark and the performance is similar to our internal fork.

Though I've check the v1.0.0  (1.0.0.BUILD-20190402.182354-150) and the performance is bad there. In the test I'm calling 1000 times an endpoint which is returning list of 1000 items with "self" link. On v1.0.0 it takes 89,855s vs 0,571s without links and ~19s with newest 0.25.x.  I'm not sure what's the cause of it. Both tests were against spring 5.1.2.RELEASE. I'll create a bug for it and include the benchmark code.

Second thing. Internally we are using faster implementation of link generation (twice faster than 0.25.x after changes). Though to get better implementation I had to change the way how the links generation works. I'll create PR for that. Maybe you will be able to include something from it.

odrotbohm:
Thanks for the feedback, @mst-avaleo. Looking forward to the PRs.

PR #751, opened 2018-11-22
Align assertions with its Javadoc in Hop.header()

izeye:
This PR aligns assertions with its Javadoc in `Hop.header()`.


gregturn:
Resolved via ca0263f.

gregturn:
Thanks @izeye !

PR #759, opened 2018-12-07
#667 - Ignore unknown properties in LinkMixin.

gregturn:
In HAL-based documents (HAL and HAL-FORMS), the "templated" property must ignored in LinkMixin, because that's where it is applied. Removed the need for a no-op `setTemplated` method in Link.

* Test cases drafted for both HAL an HAL-FORMS to verify proper behavior.
* Jackson2HalFormsIntegrationTest migrated from Hamcrest to AssertJ assertions.

Related pull requests: #668

### DIFF
--- a/src/main/java/org/springframework/hateoas/Link.java
+++ b/src/main/java/org/springframework/hateoas/Link.java
@@ -216,18 +216,6 @@ public class Link implements Serializable {
 	}
 
 	/**
-	 * This no-op setter is required to deserialize a link that contains a templated URL. It allows Jackson to
-	 * "set" the property, but since {@code templated} is a virtual property, the injected value {@code true} or
-	 * {@code false) is ignored.
-	 *
-	 * The method is kept private so no one attempts to actually use it.
-	 *
-	 * @param __ - don't care what value is passed in. The true value {@link #isTemplated()} is based upon the {@link UriTemplate}.
-	 */
-	private void setTemplated(boolean __) {
-	}
-
-	/**
 	 * Turns the current template into a {@link Link} by expanding it using the given parameters.
 	 * 
 	 * @param arguments
--- a/src/main/java/org/springframework/hateoas/hal/LinkMixin.java
+++ b/src/main/java/org/springframework/hateoas/hal/LinkMixin.java
@@ -30,7 +30,7 @@ import com.fasterxml.jackson.databind.annotation.JsonSerialize;
  * @author Oliver Gierke
  * @author Greg Turnquist
  */
-@JsonIgnoreProperties({ "rel", "media", "affordances", "template" })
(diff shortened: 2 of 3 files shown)


gregturn:
@schauder Yes, this setting would allow MANY properties that are mistyped or wrong to slip through the cracks, but that is actually a Good Thing™.

For RESTful apps, you generally want to avoid fast failing over properties not recognized. That's because those _might_ be properties added in future versions of the service that the current version of the service doesn't recognize.

When anyone uses Spring HATEOAS straight up, or through Spring Data REST, Spring HATEOAS's `@EnableHypermediaSupport` annotation actually activates a bean post processor that will find all instances of `RestTemplate` and register converters for hypermedia with these settings:

```java
private MappingJackson2HttpMessageConverter createHalConverter(ObjectMapper objectMapper, CurieProvider curieProvider,
		RelProvider relProvider, MessageSourceAccessor linkRelationMessageSource) {

	HalConfiguration halConfiguration = this.halConfiguration.getIfAvailable(() -> new HalConfiguration());

	HalHandlerInstantiator instantiator = new Jackson2HalModule.HalHandlerInstantiator(relProvider, curieProvider,
			linkRelationMessageSource, halConfiguration);

	ObjectMapper mapper = objectMapper.copy();
	mapper.disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
	mapper.registerModule(new Jackson2HalModule());
	mapper.setHandlerInstantiator(instantiator);

	MappingJackson2HttpMessageConverter converter = new TypeConstrainedMappingJackson2HttpMessageConverter(
			ResourceSupport.class);
	converter.setSupportedMediaTypes(Arrays.asList(HAL_JSON, HAL_JSON_UTF8));
	converter.setObjectMapper(mapper);

	return converter;
}
```

That's a lot, but nestled in the middle of configuring HAL support, we deliberately disable Jackson's `FAIL_ON_UNKNOWN_PROPERTIES` option.

We even petitioned Jackson to flip their default setting on this property, but they said no.

The tradeoff, which I'm sure you're smelling, is that you are disabling fast failure. For many scenarios, we prefer to fail fast, but in RESTful backward compatibility, we do NOT.

It means that someone could ALMOST have the right property, but if they don't it won't get picked up. Spring Data has near-miss algorithms to detect if someone ALMOST has the right property name in query derivation, and will spit out a message helping the user. We do not.

If you're curious why the post processor code above isn't handling "templated" property names, it's because Jackson mixins override POJO Jackson settings. A corner case mended with this PR.

gregturn:
Resolved via https://github.com/spring-projects/spring-hateoas/commit/8bf01c3ee57165d7afee0c825b562108c3fbc2cb.

Issue #764, opened 2018-12-12
Migrate build jobs to Concourse CI


gregturn:

gregturn:
Resolved via https://github.com/spring-projects/spring-hateoas/commit/7874be5b087d48fbce792102e91a6edac8666bea and https://github.com/spring-projects/spring-hateoas/commit/a5554ef2e598a9d1e56d3231e95452d760e6a35b.

PR #767, opened 2018-12-13
#764 - Introduce Concourse CI.

gregturn:
Introduces pipeline that replaces the Bambook Spring HATEOAS one which builds on master branch and deploys to artifactory.

Additionally:

* Tests against JDK 8, 11, and 12 before doing the final build and deploy.
* Also handles pull requests, running same suite of tests against mutiple JDKs.


ci/pipeline-template.yml
```diff
@@ -0,0 +1,427 @@
...
+  - Test - JDK 12
+  - Test - JDK 12 and Spring 5.0 (snapshots)
+  - Test - JDK 12 and Spring 5.1 (snapshots)
+  - Build
+- name: pull-requests
+  jobs:
+  - spring-hateoas-pull-requests
+
+jobs:
+- name: Test - JDK 8
```
schauder: The job configurations look awfully redundant.
Is there a way to reduce code duplication?
Like a parameterized template?

gregturn: I see the issue you cite. 

Ben says that YAML has something called "YAML references" that may support the concept needed here. I'll check it out.

gregturn:
@schauder 

Based on your comments about `pom.xml`, I went and dug up the original file and worked from there inside a docker container to minimize the impacts, while also working with the original "Build" job from Bamboo.

Results:

* Removed the wagon task that attempts to push to a VPN-based resource. Instead, deployment of the ZIP to artifactory will leverage Trevor's CRON job to unpack.
* Bamboo uses "install" which is very hard to command the target artifacts be sent to an alternative folder (desired for Concourse). So I switched to "deploy" which required adjusting the plugins slightly.
* maven-surefire-plugin presently breaks due to latest mods to JDK 8, so fixed that.

These changes now yield a less invasive change to `pom.xml` but still builds properly for Concourse.

gregturn:
@schauder I have also slimmed down the pipeline as much as I could figure out.

PR #781, opened 2019-01-04
Polish Link.valueOf()

izeye:
This PR polishes `Link.valueOf()` by removing `containsKey()` calls. When keys exist, duplicate lookups can be saved.

### DIFF
--- a/src/main/java/org/springframework/hateoas/Link.java
+++ b/src/main/java/org/springframework/hateoas/Link.java
@@ -320,34 +320,41 @@ public class Link implements Serializable {
 
 			Map<String, String> attributes = getAttributeMap(matcher.group(2));
 
-			if (!attributes.containsKey("rel")) {
+			String rel = attributes.get("rel");
+			if (rel == null) {
 				throw new IllegalArgumentException("Link does not provide a rel attribute!");
 			}
 
-			Link link = new Link(matcher.group(1), attributes.get("rel"));
+			Link link = new Link(matcher.group(1), rel);
 
-			if (attributes.containsKey("hreflang")) {
-				link = link.withHreflang(attributes.get("hreflang"));
+			String hreflang = attributes.get("hreflang");
+			if (hreflang != null) {
+				link = link.withHreflang(hreflang);
 			}
 
-			if (attributes.containsKey("media")) {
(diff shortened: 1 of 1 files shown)


gregturn:
@izeye I appreciate this. The code is sound, but I wanted to see what the performance gain would be so I wrote the following performance test:

```java
public class LinkPerfTest {

	static final int NUM_LOOPS = 20000;

	@Test
	public void load() {

		// warmup
		for (int i = 0; i < NUM_LOOPS; i++) {
			parseLink();
		}

		// start the clock
		long start = System.nanoTime();


		// test
		for (int i=0; i < NUM_LOOPS; i++) {
			parseLink();
		}

		// stop the clock
		long stop = System.nanoTime();

		long elapsed = stop - start;

		Duration perLoop = Duration.ofNanos(elapsed / NUM_LOOPS);

		System.out.println("Number of loops: " + NUM_LOOPS);
		System.out.println("Avg. time to parse RFC5988: " + perLoop.toNanos() + "ns");
	}

	private Link parseLink() {

		return Link.valueOf("</customer/1>;" //
			+ "rel=\"self\";" //
			+ "hreflang=\"en\";" //
			+ "media=\"pdf\";" //
			+ "title=\"pdf customer copy\";" //
			+ "type=\"portable document\";" //
			+ "deprecation=\"http://example.com/customers/deprecated\";" //
			+ "profile=\"my-profile\"");
	}

}
```

Results:

```
Number of loops: 20000
Avg. time to parse RFC5988: 24349ns
```

The current algorithm takes about 24 millionths of a second to parse a fully loaded RFC5988 link header. That's not slow. Probably because the map only has seven elements at the most. Looking up a key twice isn't very expensive with a map that small.

So I don't believe we need to sacrifice the readability of the current solution for performance in this scenario.

izeye:
@gregturn Thanks for the quick feedback! I just raised this as I thought it's generally good to avoid additional lookups whenever possible and it doesn't harm readability but if you think it does, we can close this one.

gregturn:
I prefer minimizing the number of intermediate variables as well as usage of `null`. So I don't want to introduce more of that unless it's justified by a significant performance improvement.

Issue #791, opened 2019-01-26
Support HAL name property


marcuslange:
I see the availability of all other HAL properties but name listed in the builder. Is there a reason name is not represented or am I missing something in the documentation?

gregturn:
Resolved.

Issue #787, opened 2019-01-18
HalLinkDiscoverer doesn't extract link properties other than href


lnr0626:
Given the following json:
```
{
  "_items": {
    "content": {
      "href": "http://localhost/content",
      "type": "text/plain"
    }
}
```

I would expect the following code to work:

```
assertThat(new HalLinkDiscoverer().findLinkWithRel("content"))
    .hasFieldOrPropertyWithValue("href", "http://localhost/content")
    .hasFieldOrPropertyWithValue("type", "text/plain");
```

However the HalLinkDiscoverer only extracts the href and ignores all other link properties.

gregturn:
Yup, it's right here inside `JsonPathLinkDiscoverer`:

```
	private List<Link> createLinksFrom(Object parseResult, String rel) {

		if (parseResult instanceof JSONArray) {

			JSONArray array = (JSONArray) parseResult;

			return array.stream() //
					.map(element -> new Link(element.toString(), rel)) //
					.collect(Collectors.collectingAndThen(Collectors.toList(), Collections::unmodifiableList));
		}

		return Collections.unmodifiableList(Collections.singletonList(new Link(parseResult.toString(), rel)));
	}
```

Something that got overlooked when we added new attributes to `Link`.

gregturn:
@lnr0626 BTW, it should be `_links`, not `_items`.

gregturn:
Resolved via https://github.com/spring-projects/spring-hateoas/commit/9e8e57a74ba5bc88f1e30b7690a9e07f0f6545aa.

PR #801, opened 2019-02-06
#791 - Add 'name' property to Link.

gregturn:


gregturn:
This is to provide support for https://tools.ietf.org/html/draft-kelly-json-hal-07#section-5.5.

gregturn:
Related to #787.

gregturn:
Should also support a bare String and include a pass through default. That way, the simplest pattern where you just want the href and the rel shouldn’t require an override.

PR #802, opened 2019-02-06
#787 - Return all Link attributes on discovery.

gregturn:


src/main/java/org/springframework/hateoas/core/JsonPathLinkDiscoverer.java
```diff
@@ -167,16 +159,33 @@ private JsonPath getExpression(String rel) {
 
 		if (parseResult instanceof JSONArray) {
 
-			JSONArray array = (JSONArray) parseResult;
-
-			return array.stream() //
-					.map(element -> new Link(element.toString(), rel)) //
+			return ((JSONArray) parseResult).stream() //
+					.flatMap(o -> {
```
odrotbohm: Ternary?

PR #856, opened 2019-03-06
Improve migration script

vpavic:
This PR improves migration script by adding support for `@Relation`.


odrotbohm:
Lovely, thanks, that's merged.

PR #860, opened 2019-03-08
Switch to URI.create()

gregturn:
`URI.create()` wraps `new URI()` and converts a `UriSyntaxException` into an `IllegalArgumentException`.


gregturn:
Looks like the examples can leverage `Link.toLink()` which is the same thing.

odrotbohm:
I've recently introduced `Link.toUri()` for sort of that purpose. Do you think we can move to that instead?

gregturn:
@odrotbohm Done!

gregturn:
Thanks @schauder!

PR #865, opened 2019-03-13
Stop building Spring HATEOAS Examples against 5.1

gregturn:


odrotbohm:
Can we postpone this until a consumable version of Spring 5.2 is out? I'm afraid we cut off users eager to try HATEOAS 1.0 milestones with existing Boot if we make the upgrade to 5.2 snapshots mandatory.

After that, let's turn that ticket into a general upgrade to 5.2, please.

gregturn:
Sorry. Typo in the header. This is about Spring HATEOS Examples, which got upgraded to Spring Boot 2.2, and thus breaks when built against Spring Framework 5.1.

odrotbohm:
Oh, I thought they lived in a separate repo.

gregturn:
Yes but they’re both governed by a single pipeline for convenience. I can split it up if you’d prefer.

gregturn:
Resolved via #919.

PR #887, opened 2019-03-19
#886 - Add more ref doc material.

gregturn:


odrotbohm:
I'm gonna take this from here but it would be cool if we could capture more of the essence of the PR in the commit message and ticket summary going forward.

odrotbohm:
This is polished and merged.

Issue #885, opened 2019-03-19
Re-enable default registry of WebFlux encoders and decoders


gregturn:
Pending https://github.com/spring-projects/spring-framework/issues/22612, we should re-enable the default registry of encoders/decoders used in `WebClientConfigurer`. The present configuration interferes with Spring Cloud Sleuth.

gregturn:
Duplicates #867.

Issue #867, opened 2019-03-15
WebClientConfigurer overrides exchange strategies


marcingrzejszczak:
Not sure if feature or bug ;) (Hateoas version 1.0.0.M1)

```
public class WebClientConfigurer {

    private final ObjectMapper mapper;
    private final Collection<HypermediaMappingInformation> hypermediaTypes;

    /**
     * Return a set of {@link ExchangeStrategies} driven by registered {@link HypermediaType}s.
     *
     * @return a collection of {@link Encoder}s and {@link Decoder} assembled into a {@link ExchangeStrategies}.
     */
    public ExchangeStrategies hypermediaExchangeStrategies() {

        List<Encoder<?>> encoders = new ArrayList<>();
        List<Decoder<?>> decoders = new ArrayList<>();

        this.hypermediaTypes.forEach(hypermedia -> {

            ObjectMapper objectMapper = hypermedia.configureObjectMapper(this.mapper.copy());
            MimeType[] mimeTypes = hypermedia.getMediaTypes().toArray(new MimeType[0]);

            encoders.add(new Jackson2JsonEncoder(objectMapper, mimeTypes));
            decoders.add(new Jackson2JsonDecoder(objectMapper, mimeTypes));
        });

        encoders.add(CharSequenceEncoder.allMimeTypes());
        decoders.add(StringDecoder.allMimeTypes());

        return ExchangeStrategies.builder().codecs(clientCodecConfigurer -> {

            encoders.forEach(encoder -> clientCodecConfigurer.customCodecs().encoder(encoder));
            decoders.forEach(decoder -> clientCodecConfigurer.customCodecs().decoder(decoder));

            clientCodecConfigurer.registerDefaults(false);
        }).build();
    }

    /**
     * Register the proper {@link ExchangeStrategies} for a given {@link WebClient}.
     *
     * @param webClient
     * @return mutated webClient with hypermedia support.
     */
    public WebClient registerHypermediaTypes(WebClient webClient) {
        return webClient.mutate().exchangeStrategies(hypermediaExchangeStrategies()).build();
    }
}
```

this 

```
  public WebClient registerHypermediaTypes(WebClient webClient) {
        return webClient.mutate().exchangeStrategies(hypermediaExchangeStrategies()).build();
    }
```

overrides the existing `exchangeStrategies`, which results in 

```
[...]

PR #893, opened 2019-03-21
#867 - Reenable default encoders and decoders.

gregturn:
Spring Framework 5.1 puts custom encoders and decoders in the wrong place, so Spring HATEOAS can't properly register with WebFlux. To make Spring HATEOS media types work requires overriding features that break other toolkits, like Spring Cloud Sleuth. However, in Spring Framework 5.2, this is no longer needed, so we need only disable the registry when Spring Framework 5.1 is on the classpath.

Related issues: #885, https://github.com/spring-projects/spring-framework/issues/22612

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
 
 	<groupId>org.springframework.hateoas</groupId>
 	<artifactId>spring-hateoas</artifactId>
-	<version>1.0.0.BUILD-SNAPSHOT</version>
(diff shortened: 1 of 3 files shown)


gregturn:
In light of #919, the scope on this is reduced to simply re-enabling default registry.

gregturn:
Resolved via https://github.com/spring-projects/spring-hateoas/commit/fc3767b90ba0049017753caf0d1411c57464ef79.

Issue #402, opened 2015-10-18
Provide JSON API support


sdeleuze:
I create this feature request after some discussions and thoughts that happen after the [JSON API in a Java Web Application](http://www.baeldung.com/json-api-java-spring-web-app) blog post and on the [SPR-13570](https://jira.spring.io/browse/SPR-13570).

[JSON API](http://jsonapi.org/) is a specification for how a client should request that resources be fetched or modified, and how a server should respond to those requests. Clients built around JSON API are able to take advantage of its features around efficiently caching responses, sometimes eliminating network requests entirely. [Why not use the HAL specification?](http://jsonapi.org/faq/#why-not-use-the-hal-specification) is also worth to read.

Ultimately, the goal of this feature is to provide support for the [client-side Javascript JSON API libraries](http://jsonapi.org/implementations/), like the most advanced one [Ember Data](https://github.com/emberjs/data).

I am not a Spring HATEOAS expert, but my thoughts about that is that such support may be achieved by adding JSON API format support in addition to HAL, and providing standardized `ResourceController`/`RelationshipController` Spring MVC controller interfaces that may the the Spring couterpart of [Katharsis ResourceRepository/RelationshipRepository](http://katharsis.io/docs#repositories). This last point seems quite important since JSON API is not just a format, and also [defines standardized REST endpoints](http://jsonapi.org/format/#fetching) for dealing with CRUD and relationship operation. You can have a look to [RESThub RestController](https://github.com/resthub/resthub-spring-stack/blob/master/resthub-web/resthub-web-server/src/main/java/org/resthub/web/controller/RestController.java) for an example of such controller interface.

I have not the knowledge to contribute that support alone, but I will be happy to contribute.

olivergierke:
I am all in for being able to produce JSON API compliant output for `ResourceSupport` subtypes. However, it feels like the spec suffers quite a bit from feature creep (see the aforementioned [Fetching](http://jsonapi.org/format/#fetching)) which makes implementing only the hypermedia aspects quite tricky.

I guess we can start simple and take it from there.

masterspambot:
Just to mention @sdeleuze currently most advanced JS client is [Orbit.js](https://github.com/orbitjs/orbit.js) that has adapter for ember-data. It is maintained by the guys behind JSON API.

sdeleuze:
@masterspambot Thanks for noticing, I was not aware of this one. It seems very interesting (and not specific to EmberJS).

[...]

PR #969, opened 2019-03-30
#402 - Support fo JSON API media type

atkawa7:


pivotal-issuemaster:
@atkawa7 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=969)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=969) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@atkawa7 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=969)!

gregturn:
Any tests?

atkawa7:
@gregturn Not yet :). But I will. Did you see my question regarding `Identifiable`. I wanted to find out that before I could write any tests

gregturn:
Yes. I’ll let @oliverdrotbohm address that.

toedter:
great to see this!

atkawa7:
@odrotbohm Thanks I will add the requested changes

atkawa7:
@odrotbohm @gregturn 

After implementing IdentityExtractor SPI. I realized there are couple of issues.
Suppose we had a model User 

```java
@Data
class User{
    private String id;
    private String firstName;
    private String lastName;
} 
```


then we expect the resulting json api  for the endpoint `/users/1` to be 

```json
{
    "data" :{
        "id": "1",
        "type: "users",
        "attributes": {
            "firstName": "john",
            "lastName": "doe"
        }
    }
}
```


However if we introduce an `IdentityExtractor` SPI. Then we would have something like 

```java
public interface IdentityExtractor {
    default String getId(Object entity){
        return null;
    }
    default String getType(Object object){
        return null;
    }
}

public interface IdentityExtractorProvider {
    IdentityExtractor identityExtractor();
}

public class DefaultIdentityExtractorProvider implements IdentityExtractorProvider {
    @Override
    public IdentityExtractor identityExtractor() {
        return new IdentityExtractor() {
        };
    }
}
```

In this case its difficult to dynamically ignore the (type, id) fields then we will end up with 

```json
{
    "data" :{
        "id": "1",
        "type: "users",
        "attributes": {
            "firstName": "john",
            "lastName": "doe",
            "type": "user",
            "id": "1"
        }
    }
}
```

However if we let the user extends  `Identifiable` which already has ignore annotations then we won't have that issue

```java
public interface Identifiable {
    @JsonIgnore
    String getId();

    @JsonIgnore
    String getType();
}
```

Is there any way to dynamically remove (id, type) from objects. I do not want to add another `toAttributes` method in the `IdentityExtractor` interface

atkawa7:
@odrotbohm @gregturn PING

gregturn:
When I use Jackson, it's an experiment everytime, despite using it all the time.

Several options for shaping the outputs can be found here => https://www.baeldung.com/jackson-ignore-properties-on-serialization

Think about what things can be put inside the serializer/deserializer code that would JSON API specific and hence not have to leak outside that.

niltz:
@atkawa7 any progress on this? I would be interested in helping out. Is this the most up to date code, or do you have work not yet pushed to your fork?

sodastreamssolutions:
Do you have any rough plan on when more or less to merge this ?

odrotbohm:
It definitely won't be in 1.0. There is a mechanism to add media type implementation from the outside though, so if you really fancy to give it a try, the code should be able to be packaged in a separate JAR and be consumed by Spring HATEOAS. It would also be very helpful to know in case you run into issues in that exercise.

magneticflux-:
@atkawa7 What's the status of this PR? I'm packaging it as a jar for private use and testing, but I could publish it as a repo and on Bintray and give you access if you wanted.

I assume you expect these changes to eventually be under Apache License 2.0 (the license for this repo), but you still retain copyright after signing the CLA so I don't want to redistribute your code without your approval.

magneticflux-:
@odrotbohm I'm having trouble adapting it from the outside because `HypermediaType` is an enum instead of something like `MediaType`. Specifically, I'm trying to use `@EnableHypermediaSupport` so I can use Spring Data Rest. It also just seems like Spring Data Rest is tightly coupled to `HAL` currently.

gregturn:
Per #1055, you don't actually need `@EnableHypermediaSupport` in order to register custom media types. If registered as a Spring bean in the application context, custom media types should be picked up and hence configured to generate proper hypermedia respones to MVC/WebFlux requests.

@odrotbohm and I spotted that ALL the annotation above does is translate `HypermediaTypes` -> `MediaType`, and then use that to filter against the built in media types before registering them. ANY media type simply registered directly into the application context will be applied.

gregturn:
I believe the major hold up on this PR was the lack of test cases.

MojoJojo:
Hi,

Any updates on this PR?

aadityamaheshwari:
Do we have any updates on this? or even on the progress?

jameshwh:
It looks the implementation doesn't support "relationships" (correct me if i am wrong), anyway I downloaded these source files and finally get it work to output JSON:API response. The way I wrapper the response is as below in case someone want do the same as me (or feel free to let me know if there's a better way to do so), 
```java
    @RequestMapping(value = "/mobiles",
            produces = { "application/vnd.api+json", "application/json" },
            method = RequestMethod.GET)
    public CollectionModel<EntityModel<Mobile>> getMobiles(Integer page, Integer size, String nameLike, String hardwareIdLike, String vinLike, Boolean withAssignedDriver, Integer fleetId) {
        try {
            List<Mobile> mobileList = buildMobileList();
            List<EntityModelMixin> mixins = new ArrayList<>();
            for (Mobile mobile : mobileList) {
                Link selfLink = linkTo(methodOn(MobilesApiController.class).getMobileById(""+mobile.getMobileId())).withSelfRel();
                EntityModelMixin mixin = new EntityModelMixin<>(mobile, selfLink);
                mixins.add(mixin);
            }
            Link selfLink = linkTo(methodOn(MobilesApiController.class).getMobiles(page, size, nameLike, hardwareIdLike, vinLike, withAssignedDriver, fleetId)).withSelfRel();
            return (CollectionModel) new CollectionModelMixin(mixins, selfLink);
        } catch (Exception e) {
            log.error("Couldn't serialize response for content type application/vnd.api+json", e);
            // Error to be handled somewhere
        }

        return null;
    }
```



Please remember to create a file spring.factories under the src/main/resources which is required by Spring, and put the content below in it:
```
org.springframework.hateoas.config.MediaTypeConfigurationProvider=\
 org.springframework.hateoas.mediatype.jsonapi.JsonApiMediaTypeConfigurationProvider
```
This will register the JSON API provider and be loaded accordingly.

gregturn:
See also: https://github.com/toedter/spring-hateoas-jsonapi

gregturn:
Superceded by https://github.com/toedter/spring-hateoas-jsonapi and #1337.

gregturn:
It's probably easier to let the community manage this type.

If you check in, they are probably open to contributions!

Issue #977, opened 2019-04-03
Performance improvements in link generation


mst-avaleo:
After adding changes from #749 I've check the v1.0.0 (1.0.0.BUILD-20190402.182354-150) and the performance is 3 times slower than on v0.25.x.  I also checked v1.0.0 earlier and I remember that it was slower before the changes too.

In the test I'm calling 1000 times an endpoint which is returning list of 1000 items with "self" link. On v1.0.0 it takes ~90s vs ~500ms without links and ~19s with newest 0.25.x. I'm not sure what's the cause of it. Both tests were run against spring 5.1.2.RELEASE. 

I'm attaching simple project with the benchmark so you can rerun it yourself. Unpack, and run 'mvn test' should be enough.

[spring-hateoas-tests.zip](https://github.com/spring-projects/spring-hateoas/files/3037425/spring-hateoas-tests.zip)

odrotbohm:
I think the sample code has to be significantly revamped to produce numbers that we should actually act upon. First of all, it's not a JMH benchmark which means that you're very likely to see distorted numbers driven by JIT warmups, code just being dropped by the JIT as its computed values do not escape methods etc.

Also, I think we can avoid the entire controller execution overhead as we can just declare controller types and then benchmark the direct usage of `ControllerEntityLinks`. With that, we wouldn't even have to start a Spring Container.

I just gave this a spin and and here were the numbers before I made some optimizations:

- 0.25.1.RELEASE – 18k ops/s
- 1.0.0.BUILD-SNAPSHOT – 14k ops/s

After a few tweaks I got 1.0.0.BUILD-SNAPSHOT – to 40k ops/s, which seemed to be quite neat. Most of that was driven by some optimizations in our affordance building APIs and some improved caching in `UriTemplate` and `MethodParameters`. I'll commit those changes against this ticket and will polish up the benchmark later this week so that we have a baseline to monitor the changes in the 1.0 branch.

I've also filed spring-projects/springframework#22730 as `ReflectionUtils.isObjectMethod(…)` also showed up as hotspot during the profiling session.

mst-avaleo:
Great work Oliver. Yesterday I also started looking deeper into UriTemplate.toString. Though your change is more wider.

Any motivation behind focusing only on EntityLinks? I've checked your changes on EntityLinks and the optimizations are really significant. Though when using WebMvcLinkBuilder it's better but still not acceptable on 1.0.0 (around 2x slower than on 0.25.1)

[...]

PR #978, opened 2019-04-03
Faster implementation of link generation

mst-avaleo:
I'm posting an alternative implementation for link generation which is 2x faster than the newest 0.25.x version. In the benchmark posted in #977 it gets around 7-9s.

The alternative implementation is based on the idea that link generation should be a highly cacheable operation. Links are usually referring to limited number of controller methods / resources, so there should be possibility to create some set of templates which will be just filled with parameters which are coming from the request or context. Building template should be done once, and it should contain most of the costly operations (like processing of annotations etc). Later filling the template with request parameters should be as simple as possible.

Unfortunately I haven't found a way to implement that by extending current process. The proposed change is a implementation on side, which is using building blocks from the library. 

Here's example of the usage:
```
import static org.springframework.hateoas.mvc.ControllerLinkBuilder.methodOn;
import static org.springframework.hateoas.mvc.FastLinks.linkTo;

String href = FastLinks.linkTo(methodOn(SampleController.class).sampleMethod(1L));
```

The PR is against 0.25.x since version 1.0.0 is a lot of slower (see #977).

We are using this fork for more than year in several services on production systems.


pivotal-issuemaster:
@mst-avaleo Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=978)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=978) to manually synchronize the status of this Pull Request.

odrotbohm:
Can you elaborate on where you think the performance wins are coming from mostly. I see a lot of encoding code and wonder if we wouldn't be better off fixing that in Spring Framework itself. URI encoding is a science in itself and I am not really eager to start maintaining that kind of code if there's a well maintained alternative in the Framework. If that is slow, we should fix that.

The lack of Javadoc is problematic as well.

You argue that the templates are highly cacheable, which makes me wonder whether we could actually get away with more dedicated caching in various places.

odrotbohm:
After playing with the benchmark and the profiler today I had a few ideas I thought I’d summarize here as they seem related to the proposal. The current `LinkBuilder` implementations are based on `UriComponents` to a large degree. That means the call to `linkTo(…)` immediately creates an expanded URI containing the values handed into the dummy method invocations. This approach allows us to create the `UriComponents` in one pass, which seems effective.

However, that mix up between static information (essentially a URI template backing the method) and dynamic information (the parameter values used during the dummy method invocation). That prevents us from caching the static part. If we built up that in a first pass, we could remember that based on the sole static method information and for subsequent invocations of the same method, we could reuse that template and simply expand it with different values.

It felt like your implementation was sort of heading into the same direction. I just think we shouldn’t have the need to implement as many fundamental building blocks ourselves. Would you be interested in trying to implement this two-pass algorithm based on the existing abstractions (`UriTemplate` et al.)?

mst-avaleo:
Implementing two-pass algorithm using existing abstractions will be more difficult than doing it on side which took several man days (BTW I full agree with all its cons you pointed out like reimplementation of fundamental building blocks). Plus: the latest implementation on 0.25.x branch has similar the performance to the impl in the PR (and to our internal fork).  In result I'll rather not get enough time needed for such two-pass algorithm when expected improvement will not be significant. 

Anyway I'm very happy that my PRs started a discussion regarding performance and triggered several improvements in the library :-)

PR #980, opened 2019-04-08
Migrate to JUnit 5.

gregturn:


odrotbohm:
@gregturn – Did you see my comments above? 🤔 I'd like to avoid to do the migration half way.

gregturn:
I reduced visibility of test classes and methods per https://github.com/spring-projects/spring-hateoas/commit/950a3a73dbea39fe594a77da8714e35c6dee3436.

odrotbohm:
Nice, I must have missed that. 😱 Thanks!

Issue #857, opened 2019-03-06
Use List not Collection when asking for beans


gregturn:
`List` will honor Spring’s `@Order` while `Collection` will not.

PR #1027, opened 2019-07-31
#857 - Use List instead of Collection in configuration classes.

gregturn:


gregturn:
@odrotbohm ping!

gregturn:
Resolved via 6d17d7e9.

Issue #1047, opened 2019-08-09
WebFlux configuration must properly handle application/json


gregturn:
When HATEAOS is activated, it seems to remove `application/json`, the default media type used by tools like `WebTestClient`. This can be irritating if you have controllers that do NOT deal with hypermedia yet seem to need a `Content` request header of `application/hal+json`.

gregturn:
Probably superceded by #1061.

PR #1048, opened 2019-08-09
#1047 - Properly respond to application/json calls.

gregturn:
Spring HATEAOS must be strict in what it registers, thus NOT use WebFlux's defaults. Yet it should register at least it's HAL-based media types for application/json, the default type used by testing toolkits such as WebTestClient. That way, if you craft a controller that does not deal with hypermedia, it will function as expected.

That is why I've expanded HypermediaMappingInformation to include a `getRegisterableMediaTypes` method. By default, it simply returns the results of `getMediaTypes`. But certain media types, like HAL and HAL-FORMS, can add `application/json` as a media type they will respond to if asked. This provides the means to support other formats using the same serializers.

### DIFF
--- a/src/main/java/org/springframework/hateoas/config/HypermediaMappingInformation.java
+++ b/src/main/java/org/springframework/hateoas/config/HypermediaMappingInformation.java
@@ -15,6 +15,7 @@
  */
 package org.springframework.hateoas.config;
 
+import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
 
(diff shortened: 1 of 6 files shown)


odrotbohm:
I'd love to get on a quick call with @rstoyanchev this week to discuss this approach. My first reaction towards the solution was rather conservative as I am not sure we want to expose this problem in the SPI as to me it seems like a end user decision which of the JSON-based media types is supposed to be the one to be returned for a request asking for `application/json`, not the the media type implementor.

Issue #1038, opened 2019-08-04
Update documentation to show how to create an ALPS serving controller for profile purposes.


gregturn:
Problem:

Create a web controller method that returns `Alps`. It will typically produce either `application/json` or `application/hal+json`. Change the `produces` clause to `ALPS_JSON_VALUE`, and things fall apart.

```java
@GetMapping(value = "/hypermedia/items/profile", produces = MediaTypes.ALPS_JSON_VALUE)
public Alps profile() {
	return alps() //
			.descriptor(Collections.singletonList(descriptor() //
					.id(Item.class.getSimpleName() + "-representation") //
					.descriptor(Arrays.stream(Item.class.getDeclaredFields()) //
							.map(field -> descriptor() //
									.name(field.getName()) //
									.type(Type.SEMANTIC) //
									.build()) //
							.collect(Collectors.toList())) //
					.build())) //
			.build();
}
```

A Spring REST Docs test method like this...

```java
@Test
void findProfile() {
	this.webTestClient.get().uri("/hypermedia/items/profile") //
			.accept(MediaTypes.ALPS_JSON) //
			.exchange() //
			.expectStatus().isOk() //
			.expectBody() //
			.consumeWith(document("profile", //
					preprocessResponse(prettyPrint())));
}
```

...will produce this type of stack error:

```
org.springframework.web.reactive.function.UnsupportedMediaTypeException: Content type 'application/json' not supported for bodyType=java.util.LinkedHashMap<?, ?>
	at org.springframework.web.reactive.function.BodyInserters.unsupportedError(BodyInserters.java:368) ~[spring-webflux-5.2.0.BUILD-20190802.092224-620.jar:5.2.0.BUILD-SNAPSHOT]
	at org.springframework.web.reactive.function.BodyInserters.lambda$writeWithMessageWriters$11(BodyInserters.java:358) ~[spring-webflux-5.2.0.BUILD-20190802.092224-620.jar:5.2.0.BUILD-SNAPSHOT]
	at java.util.Optional.orElseGet(Optional.java:267) ~[na:1.8.0_212]
```

gregturn:
Turns out, since there are no custom serializers/deserializers to register, there is nothing to do except possibly update documentation illustrating how to publish your own profile.

gregturn:
Thanks to #1064, the scope of this has shrunk from a bug to a simple request to update documentation and add some more tests.

gregturn:
[...]

PR #1053, opened 2019-08-13
#1038 - Update documentation for ALPS.

gregturn:


gregturn:
With #1064 merged to master, this PR has simplified to just adding test cases and documentation improvements.

src/main/java/org/springframework/hateoas/mediatype/alps/Alps.java
```diff
@@ -38,14 +40,23 @@
...
+@JsonPropertyOrder({ "version", "doc", "descriptor" })
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public class Alps {
 
-	private final String version = "1.0";
+	private final String version;
 	private final Doc doc;
 	private final List<Descriptor> descriptor;
 
+	@JsonCreator
```
odrotbohm: Can you elaborate on why we need the creators? So far the classes have been designed to only be used for serialization and object creation is handled via builders.

src/main/java/org/springframework/hateoas/mediatype/alps/Descriptor.java
```diff
@@ -33,14 +35,32 @@
...
-	private final String id, href, name;
+	private final String id;
+	private final String href;
+	private final String name;
 	private final Doc doc;
 	private final Type type;
 	private final Ext ext;
 	private final String rt;
 	private final List<Descriptor> descriptor;
+
```
odrotbohm: See above question.

src/main/java/org/springframework/hateoas/mediatype/alps/Alps.java
gregturn: It makes deserialization work which seemed necessary for using WebTestClient. 

What’s wrong with making deserialization operational? We can make that constructor private.

gregturn: I think I had test cases earlier that tapped deserialization, but I clearly changed them to use `jsonPath` in the MockMVC and WebTestClient test cases. I can remove the `@JsonCreator` if you'd like.

gregturn:
Resolved via https://github.com/spring-projects/spring-hateoas/commit/f15740fb25f7babb34f423e9a54c331cb213e43a.

Issue #1061, opened 2019-09-01
Enabling hypermedia disables all of WebFlux's default codecs


wilkinsona:
If I `@EnableHypermediaSupport`, all of the default server codes are lost:

https://github.com/spring-projects/spring-hateoas/blob/e5669eefa02d48a0143e9ab1cefa6185bc115599/src/main/java/org/springframework/hateoas/config/WebFluxHateoasConfiguration.java#L141

As far as I can tell, this makes Spring HATEOAS difficult to use in a WebFlux application that doesn't solely contain HATEOAS-based endpoints. It's also a significant departure from using HATEOAS with Spring MVC where the Web MVC customisations are additive.

gregturn:
Okay, this was originally put in because something about the order of custom registered mediatype handlers and the defaults didn't work right. I'm happy to report that appears to no longer be an issue. Flipping to `true` causes almost all test cases to pass.

The one failing test case is where `HAL` and `HAL-FORMS` are register, but a request for `UBER` is made. This was originally coded to fail, some 400 HTTP status code.

Switching to `true` causes it to return an HTTP 200 status code, and when querying the root URI of the test case:

```java
@Configuration
@EnableWebFlux
static abstract class BaseConfig {

	@Bean
	TestController testController() {
		return new TestController();
	}
}

@EnableHypermediaSupport(type = HAL)
static class HalWebFluxConfig extends BaseConfig {}
```

```java
@GetMapping
RepresentationModel<?> root() {

	RepresentationModel<?> root = new RepresentationModel<>();

	root.add(new Link("/").withSelfRel());
	root.add(new Link("/employees").withRel("employees"));

	return root;
}
```

```java
@Test
void callingForUnregisteredMediaTypeShouldFail() {

	setUp(HalWebFluxConfig.class);

	this.testClient.get().uri("/").accept(MediaTypes.UBER_JSON) //
			.exchange() //
			.expectStatus().isOk() //
			.returnResult(String.class) //
			.getResponseBody() //
			.as(StepVerifier::create) //
			.expectNextMatches(s -> {
				return true;
			}) //
			.verifyComplete();
}
```

it yields the following JSON: `{"links":[{"rel":"self","href":"/"} {"rel":"employees","href":"/employees"}]}`

This is basically falling back to no registered codec, and rendering the Spring HATEAOS `RepresentationModel` object using plain Jackson, i.e. no hypermedia format.

[...]

Issue #1052, opened 2019-08-12
Make it possible to consume Spring Boot Actuator hypermedia


gregturn:
Consuming Spring Boot Actuator’s hypermedia should be easy. Check if Boot is serving up the right media type and then verify we can consume it with:

* RestTemplate
* WebClient
* Traverson

NOTE: Recommend label “integration”.

gregturn:
Inspired by #1050 but we should target Boot 2.2.

PR #1064, opened 2019-09-04
#1061 - Reenable default codecs for WebFlux configuration.

gregturn:
Spring WebFlux now correctly handles custom codecs. This means we can stop disabling them. This change includes more test cases, verifying Spring MVC configuration for hypermedia as well, verifying both Web MVC and WebFlux are properly and consistently configured.

Probably supercedes: #1047

### DIFF
--- a/src/main/java/org/springframework/hateoas/config/WebFluxHateoasConfiguration.java
+++ b/src/main/java/org/springframework/hateoas/config/WebFluxHateoasConfiguration.java
@@ -124,21 +124,14 @@ class WebFluxHateoasConfiguration {
 		@Override
 		public void configureHttpMessageCodecs(ServerCodecConfigurer configurer) {
 
-			CodecConfigurer.CustomCodecs customCodecs = configurer.customCodecs();
-
 			this.hypermediaTypes.forEach(hypermedia -> {
 
-				MimeType[] mimeTypes = hypermedia.getMediaTypes().toArray(new MimeType[0]);
-
 				ObjectMapper objectMapper = hypermedia.configureObjectMapper(this.mapper.copy());
-				customCodecs.encoder(new Jackson2JsonEncoder(objectMapper, mimeTypes));
-				customCodecs.decoder(new Jackson2JsonDecoder(objectMapper, mimeTypes));
+				MimeType[] mimeTypes = hypermedia.getMediaTypes().toArray(new MimeType[0]);
+				
+				configurer.customCodecs().encoder(new Jackson2JsonEncoder(objectMapper, mimeTypes));
+				configurer.customCodecs().decoder(new Jackson2JsonDecoder(objectMapper, mimeTypes));
 			});
-
-			customCodecs.encoder(CharSequenceEncoder.allMimeTypes());
-			customCodecs.decoder(StringDecoder.allMimeTypes());
-
-			configurer.registerDefaults(false);
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


gregturn:
Feedback welcome @wilkinsona.

wilkinsona:
Thanks for the fix, @gregturn. +1 for leaving WebFlux's default codecs enabled as disabling them makes WebFlux very cumbersome to use for any endpoints that aren't using Spring HATEOAS.

I am less sure about the fallback to Jackson, even if it is now consistent across MVC and WebFlux. For a request that accepts, say, `application/vnd.amundsen-uber+json`, it seems wrong to respond with plain JSON that, while legal JSON, is in the wrong format for the requested media type.

I discussed this a bit with @bclozel today and he remarked that it's an interesting use case and one that Framework would like to support. It may be too late in the schedule now for Framework 5.2 and HATEOAS 1.0, but perhaps there's something that can be done without any Framework changes.

gregturn:
If be curious what the basis for the wild card matcher? Sounds like it may have been put in to catch a particular situation that is now actually an issue.

wilkinsona:
I don't know why it was added originally, but I do know it's useful for any endpoint that returns an `application/*+json` response that doesn't require any customization of Jackson to serialise correctly. Boot's Actuator is one example.

wilkinsona:
It occurred to me overnight that there have been number of Boot issues (https://github.com/spring-projects/spring-boot/issues/2147 is a good starting point) related to this where users expected a request for `application/json` to return a HAL-formatted response but it did not.  That expectation seems reasonable to me as `application/hal+json` is a form of `application/json`. The inverse, however, is not the case which is the problem with the fallback for which a test is added in the pull request.

gregturn:
I think the scope of the issue is limited because I don’t think we have a lot of people asking for, say, Collection+JSON where it wasn’t registered. Which means we can mend that in the next versions.

That being said, my inclination is to remove the wildcard matcher and to instead register handlers as needed. For Boot’s Actuator, you could register one using HATEOAS’s SPI for custom media types. 

And when HATEOAS is NOT on the classpath, well you’d have to register that same handler directly with Spring Framework.

This connects with https://github.com/spring-projects/spring-hateoas/issues/1052

wilkinsona:
I really don't think HATEOAS should do that. While we could do something in Boot's Actuator to work around the breaking change, it would also be a breaking change for any user application with similar endpoints.

bclozel:
I'm not sure I'm following the whole discussion, but the `"application/*+json"` MIME type is registered by default in Spring Framework because many applications are using a custom type for API versioning. For example:

```
curl -H "Accept: application/vnd.initializr.v2.1+json" https://start.spring.io
```

I don't think we should disable that behavior in any case.

gregturn:
Since the discussion about handling `application/**-json` is moving off topic of this one, do we need to continue it elsewhere? Or consider it resolved for now?

wilkinsona:
Let's move it elsewhere please. I don't think it can be considered resolved and something probably needs to be done before 1.0 GAs and reduces the options available to fix it.

gregturn:
Resolved via https://github.com/spring-projects/spring-hateoas/commit/fd70fd90d2fa1977c87bd707457aa8859efe8274.

gregturn:
Thanks @wilkinsona @bclozel for your feedback. This issue is now resolved regarding Spring HATEOAS.

wilkinsona:
Is there a new HATEOAS issue to discuss the problem's with `application/*+json` handling?

gregturn:
Frankly, I thought that was a Framework issue to discuss. ;)

wilkinsona:
There may be some changes needed in Framework but I think changes will also, if not exclusively, be needed in Spring HATEOAS. IMO, the problem should be tracked here until a plan of attack has been devised. At that point, a Framework issue can be opened if it's needed.

gregturn:
See #1073 for the continuation of this broader topic. Whether or not this is a Framework issue or what can be found there.

PR #1065, opened 2019-09-04
When EnableHypermediaSupport(types) is an empty array, enable ALL hypermedia types.

gregturn:
Also, default types to an empty list.


odrotbohm:
Removing the milestone as it has not been merged yet. See #1265.

Issue #1074, opened 2019-09-05
Provide a mechanism to server hypermedia responses to requests that accept application/json


wilkinsona:
In Spring Boot 1.2 (I think) we added support for serving HAL in response to requests that accept application/json. This was originally mentioned in [this comment](https://github.com/spring-projects/spring-hateoas/pull/1064#issuecomment-528227035) but I think it ought to be tracked as a separate issue.

The way that we achieve the behaviour described above has changed a bit over time. The current implementation relies on the mutability of `TypeConstrainedMappingJackson2HttpMessageConverter` to add `application/json` to its supported media types. We need to be able to do something similar for WebFlux and I don't think it's possible without resorting to reflection at the moment.

Taking a step back, perhaps it would be better to revisit the approach that we took for MVC at the same time and for Spring HATEOAS to provide a first-class configuration setting for this?

This is a similar issue to #1073 but is about allowing requests for `application/json` to receive a HAL (or other format) response.

gregturn:
On this commit I coded support where a given media type could also register for application/json.

https://github.com/spring-projects/spring-hateoas/commit/674ea29efec512f8407438d0df5443f844439b8c

It feels kind of clunky. Not sure what a proper API would be that says “when asking for plain JSON, serve HAL”. Maybe it is this?

Or perhaps this but just for HAL.

gregturn:
We have a dedicated HalConfiguration bean to alter rendering of HAL. Perhaps that could be the place to “register” additional media types for the same serializer.

gregturn:
In fact putting this setting inside HalConfiguration sounds more appealing than encoding such decision into the media type handler itself. It would also grant Boot the ability to inject this through any means desired.

wilkinsona:
Thanks for taking a look, Greg. FWIW, https://github.com/spring-projects/spring-hateoas/commit/674ea29efec512f8407438d0df5443f844439b8c doesn't look too clunky to me. One thing I did notice is that I think the default is now that HAL would served in response to a request that accepts `application/json`. I wouldn't change the default in HATEOAS, just provide an option to enable that behaviour.

> It would also grant Boot the ability to inject this through any means desired.

That's what we really need from a Boot perspective. If you have time to code something up in that direction then I'd be happy to take a look at how it would integrate into Boot. It may allow us to get https://github.com/spring-projects/spring-boot/issues/16020 into Boot 2.2.

gregturn:
wdyt @wilkinsona?

[...]

Issue #1073, opened 2019-09-05
What is a proper response when an unregistered media type is request?


gregturn:
If only `HAL` is registered via `@EnableHypermediaSupport`, and the client requests `Collection+JSON` via an `Accept` header, Spring Framework will fall back to it's handler for `application/*-json`.

Is this proper behavior? If not, what is?

This is a continuation of the discussion from #1064.

gregturn:
I’m not sure what the proper response here is. 

Is the idea is to throw an exception when asking for a hypermedia type not registered but to fallback to that catchall handler?

Maybe HATEOAS needs a handler registered after all others that if it gets hit and the requested type is hypermedia mediatype it throw an exception. But if the requested mediatype is NOT one of ours it lets it go. 

Not sure if any of this is blocking for 1.0 GA.

wilkinsona:
I'm not sure if it is either. My concern is that, depending on what the answer is, breaking changes may be required to implement it. It feels like the sort of thing that should be resolved before the restrictions of reaching 1.0 come into play.

gregturn:
When I register this `Encoder` after all others, it properly fails if you ask for an unregistered media type.

```java
	private static class NonEncoder implements Encoder<Object> {

		private final List<MimeType> hypermediaMediaTypes;

		public NonEncoder(List<MimeType> hypermediaMediaTypes) {
			this.hypermediaMediaTypes = hypermediaMediaTypes;
		}

		@Override
		public boolean canEncode(ResolvableType elementType, MimeType mimeType) {
			return mimeType == null
					|| hypermediaMediaTypes.stream().anyMatch(mediaType -> mediaType.isCompatibleWith(mimeType));
		}

		@Override
		public Flux<DataBuffer> encode(Publisher<?> inputStream, DataBufferFactory bufferFactory,
				ResolvableType elementType, MimeType mimeType, Map<String, Object> hints) {
			return Flux.error(new MediaTypeNotSupportedStatusException(mimeType + " support was never registered!"));
		}

		@Override
		public DataBuffer encodeValue(Object value, DataBufferFactory bufferFactory, ResolvableType valueType, MimeType mimeType, Map<String, Object> hints) {
			throw new MediaTypeNotSupportedStatusException(mimeType + " support was never registered!");
		}

		@Override
		public List<MimeType> getEncodableMimeTypes() {
			return hypermediaMediaTypes;
		}
	}
```

wilkinsona:
[...]

PR #1077, opened 2019-09-16
#1074 - Add option to register additional media types for HAL and HAL-FORMS.

gregturn:
Inside HalConfiguration and HalFormsConfiguration, add a "wither" to support registering additional mediatypes. This makes it possible for Spring Boot to create a bean that will yield HAL when clients ask for application/json.

NOTE: This is NOT implemented in Collection+JSON, UBER, or any other mediatypes, and has not been captured as an interface that all custom mediatypes would pick up. So far, it's custom for HAL and HAL-FORMS. If there is a big draw, we can look at extracting some type of interface. But for now, community feedback would be nice before going down that road.

### DIFF
--- a/src/main/java/org/springframework/hateoas/mediatype/hal/HalConfiguration.java
+++ b/src/main/java/org/springframework/hateoas/mediatype/hal/HalConfiguration.java
@@ -20,12 +20,15 @@ import lombok.AllArgsConstructor;
 import lombok.Getter;
 import lombok.experimental.Wither;
 
+import java.util.ArrayList;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 
 import org.springframework.hateoas.Link;
 import org.springframework.hateoas.LinkRelation;
+import org.springframework.http.MediaType;
 import org.springframework.util.AntPathMatcher;
 import org.springframework.util.Assert;
 import org.springframework.util.PathMatcher;
@@ -47,6 +50,7 @@ public class HalConfiguration {
 	 */
 	private final @Wither @Getter RenderSingleLinks renderSingleLinks;
 	private final @Wither(AccessLevel.PRIVATE) Map<String, RenderSingleLinks> singleLinksPerPattern;
+	private final @Getter List<MediaType> additionalMediaTypes = new ArrayList<>();
 
 	/**
 	 * Creates a new default {@link HalConfiguration} rendering single links as immediate sub-document.
@@ -121,5 +125,17 @@ public class HalConfiguration {
 		 * A single {@link Link} is rendered as a JSON Array.
 		 */
 		AS_ARRAY
+	}
+
+	/**
+	 * Register other {@link MediaType}s this one should response to.
+	 *
+	 * @param mediatype
+	 * @return HalFormsConfiguration with new {@link MediaType} added
+	 */
+	public HalConfiguration withAdditionalMediatype(MediaType mediatype) {
+
+		this.additionalMediaTypes.add(mediatype);
+		return this;
 	}
 }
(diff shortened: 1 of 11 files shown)


gregturn:
@wilkinsona Perhaps you'd better care to review it here?

wilkinsona:
Thanks, Greg. This seems to work nicely for Spring Boot's purposes with MVC: https://github.com/wilkinsona/spring-boot/commit/3522870afd7acb2287615404cb9c820b3de05243. Unfortunately, I have not yet had time to test it with WebFlux due to the ongoing problems there (#1080).

wilkinsona:
Any chance of this landing in snapshots soon? If it lands at the last minute, the couple of days between HATEOAS's release and Boot's release is likely to be filled with work picking up everyone's releases and the updates to Boot's HATEOS auto-configuration won't make it in.

odrotbohm:
As indicated in the Slack conversation, I don’t think that this should be resolved on the HMC/Codec level as without further changes in MVC/WebFlux, their defaulting might cause invalid representations selected. 

If – as suggested – the media type aliasing is handled by the core framework we don’t need Boot to reach out to all HMCs and tweak their configuration and the HMC implementations don’t need to expose API to do that in the first place.

wilkinsona:
Sorry, I missed that in the Slack conversation.

> their defaulting might cause invalid representations

I don't think I understand why this is the case. What circumstances will lead to an invalid representation being served?

What I'd like to be able to do is to have `application/hal+json` served in response to a request that accepts `application/json`. That seems reasonable to me as HAL is JSON just in a specific format. Perhaps the solution here is too broad as it allows any media type to be added and that's what you mean by creating the possibility of an invalid representation being served?

We're a bit stuck between a rock and a hard place now. Without some sort of answer for #1074, I think we're going to run out of time and auto-configuration for Spring HATEOAS's WebFlux won't make it into Boot 2.2. The alternative is to rework the `spring.hateoas.use-hal-as-default-json-media-type` property to make it clear that it's servlet specific. Unfortunately, I've already spent all the time I had available and then some on getting to this point so that's likely to be a step too far.

gregturn:
I don’t know. Seems like a simple option that allows Boot support.

. If/when Framework offers a better option we can deprecate this
. It doesn’t break anyone currently since it’s an opt-in choice.
. Pushes things forward in a positive way

I don’t want waiting for perfect to obstruct us from something that’s good.

wilkinsona:
I am inclined to agree, @gregturn. Your proposed changes don't do anything by default in Spring HATEOAS and Boot will have a property to control its behaviour. In the absence of a change in Framework – which isn't going to happen at this late stage in 5.2's development – this proposal or something similar feels like a reasonable and pragmatic compromise to me.

odrotbohm:
Again, I am not opposed to introduce the suggested changes *if* we end up deciding that that's necessary. After the Slack discussion, I don't think it is. And I don't think we should introduce API we might have to deprecate and eventually remove if it isn't necessary.

As far as I understood, there's some defaulting in Spring Framework, that adds a catch all path to requests for mediatypes matching `application/*+json` . Now imagine we have HAL FORMS (`application/prs.hal-forms+json`) registered but the client asks for Collection+JSON (`application/vnd.collection+json`). HAL Forms wouldn't be a direct match, but the fallback to a simple match to `application/*+json` would still select it and return HAL FORMS when the client asked for Collection JSON.

I think a first step we should take is to decide where the discussion is actually supposed to be held because we're now between slack, a ticket and a PR and the conversation just gets dropped in the middle of it and is then taken over by others in a different spot so that there's quite a bit of work and overhead induced by reiteration, re-discussion. We shouldn't try to resolve this with with a "Oh yeah, let's merge this real quick.". I'm pretty sure an approach like that would never be an option for Boot itself. I don't see why we should act any different here.

Let's please step back to discuss #1074 and #1073 as they seem very related and get back to discuss implementation approaches *when* we have agreed on whether Spring HATEOAS is the place that actually is the place where the fix needs to be implemented.

Issue #1098, opened 2019-10-21
WebMvcHateoasConfiguration not applied for plain client application


jvalkeal:
This is for hateoas 1.0.0. As a simple client wanting to consume hateoas content via traverson we need to have a specific HAL module loaded for `RestClient`.

Issue seem to be in a `WebStackImportSelector` which uses `WebStack` to see if, well, stack is available. As those will check classpath if `org.springframework.web.servlet.DispatcherServlet` or `org.springframework.web.reactive.DispatcherHandler`, conditions are never true in a client app having no server side classes.

Thus `@EnableHypermediaSupport(type = EnableHypermediaSupport.HypermediaType.HAL)` on a client side don't enable needed jackson configs.

gregturn:
@odrotbohm I've coded a solution that splits apart client-side beans from server-side beans, so in a client-only setup (only having `spring-web` on the classpath), Spring HATEOAS still operates as expected and properly wires `RestTemplate` beans.

However, this makes the Java enum `WebStack` not sound right. It was aimed at being either **Spring MVC** or **Spring WebFlux**. That has been widened to really mean **server side Spring MVC**, **server side Spring WebFlux**, **client side Spring MVC**, or **client side Spring WebFlux**. And those are NOT web stacks.

So we need something different to name it. And we probably need to thus deprecate `WebStack`.

gregturn:
Resolved via 03ac8021.

gregturn:
@odrotbohm Is this fix a candidate for backporting to `1.0.x`? If the classes are package private, then this shouldn't break existing code, right?

odrotbohm:
Surely is, yes!

PR #1099, opened 2019-10-21
#1098 - Institute checks for client-only setups.

gregturn:


odrotbohm:
I like where the configuration implementation changes go but am pretty sure we should find a different way to expose that in the configuration. As you already described in the ticket, the client/server distinction does not map very well to the `WebStack` enum.

I wonder if we can hide the the distinction between simple classpath checks rather. If both the server and client related configuration are available as separate configuration classes, we could still guard their inclusion by classpath checks for e.g. `DispatcherServlet` or a simple test for `RestTemplate`'s presence.

I know that technically, `RestTemplate` is not part of MVC (so that activating support for it is a bit weird) but the original split we were trying to establish with the enum was servlet/blocking VS. reactive/non-blocking. Maybe we can introduce new enum values that better reflect this distinction in a separate ticket and deprecate the current ones. However, the need to explicitly use these flags should be an exception, not the rule.

gregturn:
@odrotbohm I've made the suggested changes. This led me to push the logic of what config files to offer into the enum as well, simplifying `WebStackImportSelector`.

My remaining question is: do we still need a public API for `isClientAvailable` and `isServerAvailable`, or can we simply keep those as private, internal variables that feed `getAvailableConfigurations()`?

On one hand, I'd say reduce the API footprint. On the other, it may be handy to know the details of your current situation.

odrotbohm:
Looks good to me! I think we can get rid of the accessors for the `boolean` flags.

gregturn:
Resolved via cf0c606e.

PR #1101, opened 2019-10-23
Fix also HAL_JSON_UTF8 media type

luvarqpp:
org.springframework.hateoas.MediaTypes.HAL_JSON_UTF8 value was removed and only HAL_JSON is present. UTF8 is considered as default encoding for application/hal+json type.


gregturn:
Resolved via f45aabe8.

gregturn:
Backported to `1.0.x`.

gregturn:
Thanks @luvarqpp!

Issue #1119, opened 2019-11-09
Configuring RestTemplate breaks in a pure Spring WebFlux application


gregturn:
While `RestTemplate` and `HttpMessageConverter` are in `spring-web`, `WebMvcConfigurer` is in `spring-webmvc`.

Spring HATEOAS leverages `WebMvcConfigurer` to help configure `RestTemplate` beans and thus breaks in a pure Spring WebFlux application since this class is unavailable.

gregturn:
NOTE: This should backported to `1.0.x` as well.

gregturn:
I just verified this patch works in my pure Spring WebFlux application against Spring Boot 2.2.1.RELEASE.

odrotbohm:
I've slightly reworked the changes and merged them. Revisiting the changes I wondered though why the class to be fixed (`RestTemplateHateoasConfiguration`) would actually be included in a bootstrap in a pure WebFlux arrangement in the first place? `WebStack` checks for the presence of `DispatcherServlet` which *is* part of `spring-webmvc` and only includes the configuration class if that's present. If it is, using `WebMvcConfigurer` should be safe in the first place, shouldn't it?

odrotbohm:
Nevermind, I got it. The previously offending configuration is included in case `RestTemplate` is on the classpath, which lives in `spring-web`. All good.

gregturn:
Problem still exists despite this patch (albeit smaller scope).

Because `org.springframework.web.client.RestTemplate` is in `spring-web`, and visible in WebFlux application, activates `RestTemplateHateoasConfiguration`.

Hence, your pure WebFlux application will invoke this:

```
@Bean
static HypermediaRestTemplateBeanPostProcessor hypermediaRestTemplateBeanPostProcessor(
		WebMvcConverters converters) {
	return new HypermediaRestTemplateBeanPostProcessor(converters);
}
```
...which fails because `WebMvcConverters` was from `WebMvcHateoasConfiguration` (as expected).

It looks like if I add:

```
	@Bean
	WebMvcConverters hypermediaWebMvcConverters(ObjectProvider<ObjectMapper> mapper,
			List<HypermediaMappingInformation> information) {
		return WebMvcConverters.of(mapper.getIfUnique(ObjectMapper::new), information);
	}
```
...to `WebFluxHateoasConfiguration`, then everything passes.

Installed in my local maven repo, my pure WebFlux app works.

[...]

PR #1120, opened 2019-11-09
#1119 - Fix RestTemplate hypermedia configuration so it doesn't use WebMvcConfigurer.

gregturn:


gregturn:
I verified this patch works in my pure Spring WebFlux application that happens to also use Spring REST Docs for testing.

odrotbohm:
That's merged and backported.

PR #1215, opened 2020-03-09
Update RFC-5988 references to RFC-8288

vpavic:
This PR updates [RFC-5988](https://tools.ietf.org/html/rfc5988) references to [RFC-8288](https://tools.ietf.org/html/rfc8288) since the former has been obsoleted by latter for around 2.5 years now.

Initially, I've split this to 3 commits that update references in docs, javadoc and code - I assume these will be merged as one commit after review, but in case some are to be handled differently (docs/javadoc vs code?) I decided to keep them apart initially.


gregturn:
Thanks @vpavic! I'll check it out.

gregturn:
Resolved via 1f493606.

gregturn:
Thanks @vpavic!

Issue #1223, opened 2020-03-23
Provide API for users to easily configure RestTemplate


gregturn:
The code originally submitted in https://github.com/spring-projects/spring-boot/pull/20373 should serve as a guide on providing an API for end users to configure hypermedia in `RestTemplate` instances.

Simplest thing could be making `WebConverters` public and then providing reference documentation on how users could do it by hand, or if they were using Spring Boot to plug it into a `RestTemplateCustomizer` to support the `RestTemplateBuilder` paradigm.

gregturn:
The simplest thing is to simply alter the visibility of `WebConverters` so end users have access to it. And provide documentation illustrating how to use it.

The quirky thing about it being a bean in `HateoasConfiguration`, is that it's need unconditionally in Spring MVC apps for server side configuration. But for a WebFlux app, it's ONLY needed if the user also wants a `RestTemplate`. Maybe if we need to keep it there, it should be a lazy bean?

PR #1238, opened 2020-03-25
#1223 - Provide API to configure RestTemplate with hypermedia.

gregturn:
By exposing WebConverters, users are able to get a hold of the registered bean and register message converters with RestTemplate instances.

Update reference docs to show different ways to configure a RestTemplate, with and without Spring Boot.


odrotbohm:
Oh, and I'd love us to get to perceived API usage symmetry with the reactive side of things. I.e. for that `WebClientCustomizer` exists in Spring Boot and the declaration of a bean that applies our customizations should look similarly simple (probably affects #1242).

gregturn:
Agreed.

That API was indeed clunky. Perhaps after the dust settles, we can retire it in favor of this one.

PR #1239, opened 2020-03-25
Provider API to configure WebClient instances.

gregturn:


gregturn:
@odrotbohm Check it out. I have actually two ways in this one. Not sure which is preferable.

Opinions @wilkinsona?

PR #1242, opened 2020-03-27
Provide API to configure WebTestClient instance.

gregturn:


pom.xml
```diff
@@ -860,7 +860,6 @@
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-test</artifactId>
 			<version>${spring.version}</version>
-			<scope>test</scope>
```
odrotbohm: Is that by accident?

src/main/java/org/springframework/hateoas/config/WebTestClientConfigurer.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.List;
+import java.util.function.Consumer;
+
+/**
+ * Assembles {@link ExchangeStrategies} needed to wire a {@link WebTestClient} with hypermedia support.
+ *
+ * @author Greg Turnquist
+ * @since 1.1
+ */
+@Configuration
```
odrotbohm: Does that need to be an `@Configuration` class? It's not exposing any beans, right?

src/main/java/org/springframework/hateoas/config/WebTestClientConfigurer.java
```diff
@@ -0,0 +1,76 @@
...
+	private Consumer<ClientCodecConfigurer> configurer;
+
+	/**
+	 * Creates a new {@link WebTestClientConfigurer} for the given {@link ObjectMapper} and
+	 * {@link HypermediaMappingInformation}s.
+	 *
+	 * @param mapper must not be {@literal null}.
+	 * @param hypermediaTypes must not be {@literal null}.
+	 */
+	public WebTestClientConfigurer(ObjectMapper mapper, List<HypermediaMappingInformation> hypermediaTypes) {
```
odrotbohm: Should this be package protected to make sure it's only us who create instances of that type?

src/test/java/org/springframework/hateoas/config/WebTestClientConfigurerTest.java
```diff
@@ -0,0 +1,166 @@
...
+import org.springframework.web.reactive.function.client.ExchangeStrategies;
+import org.springframework.web.reactive.function.client.WebClient;
+
+import java.util.Collections;
+
+import static org.assertj.core.api.Assertions.*;
+import static org.springframework.hateoas.MediaTypes.*;
+import static org.springframework.hateoas.support.ContextTester.*;
+
+public class WebTestClientConfigurerTest {
```
odrotbohm: Doesn't need to be public either, does it?

wilkinsona:
I think it would be more idiomatic, both from a Framework perspective and a Boot perspective, to work with a `WebTestClient.Builder` rather than using `mutate()` to go from a `WebTestClient` back to a `WebTestClient.Builder`. In Boot, working with the `Builder` would then allow you to configure a `WebTestClientBuilderCustomizer` bean that can apply the customization centrally rather than doing it in each test.

src/main/java/org/springframework/hateoas/config/WebTestClientConfigurer.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.function.Consumer;
+
+/**
+ * Assembles {@link ExchangeStrategies} needed to wire a {@link WebTestClient} with hypermedia support.
+ *
+ * @author Greg Turnquist
+ * @since 1.1
+ */
+@Configuration
+public class WebTestClientConfigurer {
```
wilkinsona: Framework already has `org.springframework.test.web.reactive.server.WebTestClientConfigurer`. To avoid a clash, I think a HATEOS-specific class name would be good here. Perhaps it could implement Framework's `WebTestClientConfigurer` which would allow it to be passed to `org.springframework.test.web.reactive.server.WebTestClient.Builder.apply(WebTestClientConfigurer)`?

odrotbohm:
Noted, I think we're gonna shoot for API symmetry for our customizers so that they can be turned into the Boot specific abstractions as suggested [here](https://github.com/spring-projects/spring-hateoas/pull/1238#pullrequestreview-383850333).

pom.xml
gregturn: No. We can not write `src/main/java` customizing code against `WebTestClient` if `WebTestClient` is test-scoped.

src/main/java/org/springframework/hateoas/config/WebTestClientConfigurer.java
gregturn: No doesn't need that.

src/main/java/org/springframework/hateoas/config/WebTestClientConfigurer.java
gregturn: Yes it should. Good catch.

src/main/java/org/springframework/hateoas/config/WebTestClientConfigurer.java
gregturn: Ahh. Thanks Andy. I'll look at that. We often mark stuff up like `HypemediaXXXX`, so perhaps `HypermediaWebTestClientConfigurer` would do.

Let me play around with that extension point and see if I can morph it suitably.

pom.xml
odrotbohm: Ah, that's unfortunate. We should make the dependency optional then because otherwise we always leak the test JAR into users production classpaths.

gregturn:
I’m swapping out the `registerHypermediaTypes` for Framework’s `WebTestClientConfigurer` API. Retooling the test cases. 

That should minimize our surface API.

pom.xml
gregturn: Will do.

gregturn:
Because the new class uses Reactor types, I had to alter `ArchitectureTest`.

Waiting on CI to verify it all passes. Retooled things to leverage Framework's `WebTestClientConfigurer`. If it's all green, I can merge and then carry forward to `WebClient`.

PR #1271, opened 2020-04-14
Fix #1270

### LINKED ISSUE #1270: Add Kotlin source to released artifacts
Hello,

I notice that the Koltin sources of spring hateoas are not present in the generate jar during the  maven phase attach-sources phases.

You will find soon a PR to solve this. Hope this may help.

Patouche:
This PR will add the kotlin source in the `*-sources.jar` file.

You can check the result with a command like : `unzip -l target/*-sources.jar | grep 'kt'`


gregturn:
That jar file shouldn’t be in the PR. Do you need to adjust the .gitignore file?

gregturn:
Resolved via 8995ba804a0d7eae9bc3aaa19738360013181203

Issue #864, opened 2019-03-12
Create a dedicated builder for HAL representations


gregturn:
@drdamour I'm experimenting on a fluent API to build `RepresentationModel` objects. What do you think of something like this for building an embedded structure like you depicted up above?

```java
@RestController
static class ProductController {

	LinkRelation favoriteProducts = LinkRelation.of("favorite products");
	LinkRelation purchasedProducts = LinkRelation.of("purchased products");

	@GetMapping("/products")
	public RepresentationModel<?> all() {

		EmbeddedModelBuilder builder = ModelBuilder //
				.embed() //
				.rootLink(linkTo(methodOn(ProductController.class).all()).withSelfRel());

		PRODUCTS.keySet().stream() //
				.map(id -> new EntityModel<>(PRODUCTS.get(id), new Link("http://localhost/products/{id}").expand(id))) //
				.forEach(productEntityModel -> {

					if (productEntityModel.getContent().isFavorite()) {

						builder //
								.embed(favoriteProducts) //
								.entityModel(productEntityModel) //
								.rootLink(productEntityModel.getRequiredLink(SELF).withRel(favoriteProducts));
					}

					if (productEntityModel.getContent().isPurchased()) {

						builder //
								.embed(purchasedProducts) //
								.entityModel(productEntityModel) //
								.rootLink(productEntityModel.getRequiredLink(SELF).withRel(purchasedProducts));
					}
				});

		return builder.build();
	}
}
```

It yields:

```json
{
  "_embedded" : {
    "favorite products" : [ {
      "someProductProperty" : "someValue",
      "_links" : {
        "self" : {
          "href" : "http://localhost/products/777"
        }
      }
    }, {
      "someProductProperty" : "someValue",
      "_links" : {
        "self" : {
          "href" : "http://localhost/products/998"
        }
      }
    } ],
    "purchased products" : [ {
      "someProductProperty" : "someValue",
      "_links" : {
        "self" : {
          "href" : "http://localhost/products/111"
        }
      }
    }, {
      "someProductProperty" : "someValue",
      "_links" : {
        "self" : {
          "href" : "http://localhost/products/222"
        }
      }
    }, {
      "someProductProperty" : "someValue",
      "_links" : {
        "self" : {
          "href" : "http://localhost/products/333"
        }
      }
    }, {
      "someProductProperty" : "someValue",
      "_links" : {
        "self" : {
          "href" : "http://localhost/products/444"
        }
      }
    }, {
[...]

Issue #193, opened 2014-06-05
Support for Resources in a Resource to be added to _embedded


tombee:
After speaking with @olivergierke, we came to the conclusion that there's no way to get a Resources collection added to an _embedded collection in the root of the envelope without using @JsonUnwrapped.

The problem then is that if there are multiple Resources fields, the last field serialized will overwrite the existing _embedded section.

I believe that @olivergierke suggests that a new custom serializer needs to be written to support adding Resources fields to the _embedded section.

olivergierke:
Was this closed by accident?

tombee:
Ah I thought it wasn't being addressed, since it's been open for so long without a reply :)

olivergierke:
If we finally decide not to implement it, we can still close it. Currently I am just a bit swamped with other work. :)

gregturn:
Related to #864

gregturn:
In #864, I am working on a builder that supports:

```java
@GetMapping("/explicit-and-implicit-rels")
RepresentationModel<?> explicitAndImplicitRelations() {

    Staff staff1 = new Staff("Frodo Baggins", "ring bearer");
    Staff staff2 = new Staff("Bilbo Baggins", "burglar");

    Product product1 = new Product("ring of power", 999.99);
    Product product2 = new Product("Saruman's staff", 9.99);

    return ModelBuilder3 //
            .entity(staff1) //
            .entity(staff2) //
            .entity(product1) //
            .entity(product2) //
            .link(Link.of("/people/alan-watts")) //
            .entity(LinkRelation.of("ring bearers"), staff1) //
            .entity(LinkRelation.of("burglars"), staff2) //
            .link(Link.of("/people/frodo-baggins", LinkRelation.of("frodo"))) //
            .build();
}
```

Yielding...

```
{
  "_embedded" : {
    "burglars" : {
      "name" : "Bilbo Baggins",
      "role" : "burglar"
    },
    "staffs" : [ {
      "name" : "Frodo Baggins",
      "role" : "ring bearer"
    }, {
      "name" : "Bilbo Baggins",
      "role" : "burglar"
    } ],
    "ring bearers" : {
      "name" : "Frodo Baggins",
      "role" : "ring bearer"
    },
    "products" : [ {
      "name" : "ring of power",
      "price" : 999.99
    }, {
      "name" : "Saruman's staff",
      "price" : 9.99
    } ]
  },
  "_links" : {
    "self" : {
[...]

PR #1273, opened 2020-04-17
Create builder to simplify assembling representations.

gregturn:
Original issue: #864.
Related issue: #193.


gregturn:
@odrotbohm 

I retooled things into `Model.Builder` based upon how `WebClient.Builder` is put together. I just put the `DefaultModelBuilder` and `HalModelBuilder` into the `Model` class. Then I topped it off with a `builder()` static helper as well as a `hal()` static helper.

I rewrote the test cases to get rid of MockMVC etc., and put ALL the related test cases into one class, so you can see them at a glance.

PR #1284, opened 2020-04-30
Introduce HalFormsConfiguration API to sort properties.

gregturn:


odrotbohm (changes requested):
I'd like to avoid the exposure of the type on `InputPayloadMetadata`. What if we made the `HalFormsConfiguration` expose something that exposes a `Collector` per type, so that the `PayloadMetadata` implementation based on a type could look up the property order and sort the `Stream<PropertyMetadata>` returned from `PayloadMetadata.stream()` accordingly?

PR #1305, opened 2020-06-04
Properly handle custom mediatypes registered through spring.factories.

gregturn:


gregturn:
Resolved instead via https://github.com/spring-projects/spring-hateoas/commit/4720e563f0b28f189b0885200362cfbc2b3c39d0 by updating the reference docs, indicating custom media types ONLY require implementing the `HypermediaMappingInformation`.

PR #1319, opened 2020-07-13
Wording changes

Buzzardo:
Replacing some terms


odrotbohm:
That's merged. Thanks, Jay!

PR #1353, opened 2020-08-11
Only render Map entries once.

gregturn:


odrotbohm:
I fear it's slightly more complicated than anticipated. We cannot change `getContent()` to return `null` as that – to programatic clients – has to provide access to the `Map` content as well. Otherwise the clients have to know whether they deal with an `EntityModel` that contains a `Map` or an arbitrary object, which completely defeats the purpose.

I'll investigate what Jackson customizations we can use to teach it to drop `Map`s for that property. Looks doable but it seems like we're going to have to create a Jackson dependency from within the model package which is a bit meeh.

gregturn:
Do we need an alternative Jackson rendering method, like `getNonMapContent()`, and simply make `getModel` ignored by Jackson?

odrotbohm:
Annotations are a different beast as they do not cause runtime errors if missing from the classpath. Types extending Jackson types, do, though.

I'm afraid we can't tweak the property exposure in an extensive way as media type implementations might register mixins for the `content` property and us moving to a different property would break those.

odrotbohm:
I think I've got something. A bit involved but should do the trick.

PR #1363, opened 2020-09-13
Jackson uses platform line endings.

arthur-noseda:
When building the project on Windows, I encountered an issue with one of Jackson2HalFormsIntegrationTest's methods. It seems Jackson is using the underlying platform's line ending when serializing. The expected is not met on Windows (nor should it on Mac).


cuspymd (approved):
It makes sense.

gregturn:
We HEAVILY test against Jackson. How is the ONLY place this manifests?

ParkerM:
> We HEAVILY test against Jackson. How is the ONLY place this manifests?

It looks like this is the only test that explicitly declares a JSON string containing line separators. If formatting is important for that assertion then it could use AssertJ's conventional `assertThat(serialized).isEqualToNormalizingNewlines(expected);` to fix the build on Windows.

PR #1375, opened 2020-10-04
Property Resolving in @RequestMapping for Spring Boot

lyca:
When I provided the solution for #361 I had a special use case that got solved. I thought, the solution would help also in Spring Boot, but I was wrong -> https://stackoverflow.com/questions/22167912/contextloader-getcurrentwebapplicationcontext-always-returns-null

So I worked on another solution for the problem and open this PR to discuss this one.
- should work in all Spring Boot web applications
- EnvironmentContextFilter does not need its own configuration class, I did not know, where to put it


didien:
Any chance of merging this PR for a future release, as it would fix a regression when using Spring Boot ? (please see referenced issue)

lyca:
Created an easier solution in #1411

PR #1379, opened 2020-10-19
RepresentationModelProcessor invoked for invalid CollectionModel payload if specialized

KarinaPleskach:
Hello my name is Karina and I am working on a project that uses Spring Hateoas. After migrating to Spring Boot 2.3.3, we ran into the problem that `RepresentationModelProcessor` performs a process for all objects of type `CollectionModel<EntityModel<no matter what type>>`. This is due to the fact that we created our inheritor from `CollectionModel`. The code describing the problem can be found [here](https://stackoverflow.com/questions/64333052/use-custom-inheritor-from-the-hateoas-collectionmodel-in-hateoas-server-represen).

After analyzing your code, I found out that the `RepresentationModelProcessorInvoker.getSuperType(…)` method does not handle inheritors in any way. As a result, this method returns `CollectionModelProcessorMyInheritor<?>`. Because of this, in the `RepresentationModelProcessorInvoker.isValueTypeMatch(…)`, the `resourceType` will be equal to `?`. Further, this `?`, comparing with anything in the `RMPI.isValueTypeMatch(…)` method, will return `true`, which is incorrect.

I propose a very small fix in the `….getSuperType(…)` method that consider inheritors.

I also found a change that resulted in incorrect work, since everything worked correctly before. In the `RMPI.isValueTypeMatch(…)` method, a line has appeared (highlighted in the picture) that just compares everything with `?`.
image

### DIFF
--- a/src/main/java/org/springframework/hateoas/server/mvc/RepresentationModelProcessorInvoker.java
+++ b/src/main/java/org/springframework/hateoas/server/mvc/RepresentationModelProcessorInvoker.java
@@ -434,7 +434,7 @@ public class RepresentationModelProcessorInvoker {
 
 			Class<?> rawType = source.getRawClass();
 
-			if (rawType != null && rawType.isAssignableFrom(superType)) {
+			if (rawType != null && rawType.equals(superType)) {
 				return source;
 			}
 
--- a/src/main/java/org/springframework/hateoas/server/mvc/RepresentationModelProcessorInvoker.java
+++ b/src/main/java/org/springframework/hateoas/server/mvc/RepresentationModelProcessorInvoker.java
@@ -434,7 +434,7 @@ public class RepresentationModelProcessorInvoker {
 
 			Class<?> rawType = source.getRawClass();
 
-			if (rawType != null && rawType.equals(superType)) {
+			if (rawType != null && rawType.isAssignableFrom(superType)) {
 				return source;
 			}
 
--- a/src/test/java/org/springframework/hateoas/server/mvc/RepresentationModelProcessorInvokerUnitTests.java
+++ b/src/test/java/org/springframework/hateoas/server/mvc/RepresentationModelProcessorInvokerUnitTests.java
@@ -15,12 +15,16 @@
  */
 package org.springframework.hateoas.server.mvc;
 
+import static java.util.Collections.*;
 import static org.assertj.core.api.Assertions.*;
 
-import java.util.Collections;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 import org.junit.jupiter.api.Test;
 import org.springframework.hateoas.CollectionModel;
+import org.springframework.hateoas.EntityModel;
 import org.springframework.hateoas.RepresentationModel;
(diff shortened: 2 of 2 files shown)


pivotal-issuemaster:
@KarinaPleskach Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=1379)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=1379) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@KarinaPleskach Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=1379)!

odrotbohm:
I think the change you suggest would break the contract of the `getSuperType(…)` method as it is defined. It takes a `ResolvableType` reference and a `Class` defining the super type to be resolved. We can only return the given `ResolvableType` as the super class' `RT` if it's exactly the same type, hence the `….equals(…)`. Otherwise, we'd return a `ResolvableType` instance for a `String` with a given super class `Object`, which is clearly wrong.

I just pushed a test case that's replicating what you're describing in the linked StackOverflow post (a `RMP` implementation typed to `CollectionModel<EntityModel<Foo>>` but that allegedly being invoked for a `CollectionModel<EntityModel<Bar>>`) but that test verifies what you describe does not actually happen.

Would you mind trying to tweak this test into the failure that you describe? Alternatively, please add test cases to the example code you link to so that we can reproduce the issue. So far I can't actually get to anything that's not working as expected.

KarinaPleskach:
I've added a test that shows that the wrong handler is being called. I have a single [controller](https://github.com/KarinaPleskach/Hateoas/blob/master/src/main/java/com/example/demo/Controller.java) that returns a `CollectionModel<EntityModel<DemoEntity>>`. In the [configuration class](https://github.com/KarinaPleskach/Hateoas/blob/master/src/main/java/com/example/demo/SpringDataRestConfig.java), I have `RepresentationModelProcessor` registered for both `DemoEntity` and `SecondEntity`. I added logging there and variables to count the number of calls to these methods. In the [test](https://github.com/KarinaPleskach/Hateoas/blob/master/src/test/java/com/example/demo/DemoApplicationControllerTest.java), I just check the values of these variables, which do not converge with the expected. Also, using logging, we can see that the wrong method `process(CollectionModel<EntityModel<SecondEntity>> collection)`  is called.
image
The error occurs because I am using my implementation for [CollectionModel](https://github.com/KarinaPleskach/Hateoas/blob/master/src/main/java/com/example/demo/MyHateoasCollectionModel.java). Previously, this method was not called, but after migration to spring boot 2.3.3 it is called. This is due to the fact that, as I wrote above, a comparison with the `?` was added.

odrotbohm:
Thanks for the additional test case. It helped me diagnose the original problem. I have a fix coming in a minute.

odrotbohm:
That's fixed and will be available in the snapshots shortly. Backported into 1.2.x and 1.1.x. Thanks again for you help diagnosing!

PR #1381, opened 2020-10-23
Optimized version of LinkBuilder.slash() method for identifiers

mst-avaleo:
Hi,

I'm looking on performance of following code, since it's used heavily on collection resources for `self` links in our projects, 
especially when collection is large. 

```
entityLinks.linkForItemResource(Person.class, personId).withSelfRel();
```

I've found that `slash()` method used behind the scene is a quite generic method, which can be used together with URL parts with many path segments and fragments (for example `.slash("/segment/segment2#fragment")`). Therefore there's a heavy URL processing in slash() methods used for this case. Though for the identifiers case, it's completely not needed in my opinion. Maybe I'm missing something...

In the PR, I've proposed a solution, where there's a specific `slashId()` method meant to be used only for identifiers. This enables to use simpler approach, where the input data doesn't need to be parsed from URL perspective. In result, the change gives around 20% of optimization when testing on `EntityLinks.linkForItemResource` level.


pivotal-issuemaster:
@mst-avaleo Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=1381)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=1381) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@mst-avaleo Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-hateoas&pullRequestId=1381)!

PR #1328, opened 2020-07-27
Implemented property resolving in HATEOAS links.

lyca:
#361 Added the resolving of properties by using org.springframework.web.context.ContextLoader.getCurrentWebApplicationContext().getEnvironment() for obtaining an Environment statically. This works here because the code runs in the context of a web application.
The only breaking change I see is, that testing is only possible if the ContextLoader is initialized otherwise there will be a NPE.
I am unsure, if this use case needs additional effort.

PR #1411, opened 2020-12-10
Simple solution to provide the environment.

### LINKED ISSUE #1406: Unable to resolve place holders in @RequestMapping
Using spring-hateoas:1.2.1 with Spring Boot 2.4.0:

When a REST controller class annotated with `@RequestMapping("${some-variable}/resources")` returns a resource extending `RepresentationModel`, I get the following error :
```
{
"label": "uncatched exception",
"detailMessage": "Not enough variable values available to expand 'some-variable'",
"statusCode": 500
}
```

I pinned the issue to this suspect line in the core library : https://github.com/spring-projects/spring-hateoas/blob/646e73c8c5ca5f1302bd6722e0ad338be0309fc9/src/main/java/org/springframework/hateoas/server/core/PropertyResolvingMappingDiscoverer.java#L90

From what I could read, `ContextLoader.getCurrentWebApplicationContext()` doesn't return the expected context when using Spring Boot, and it's actually returning `null` when I place a breakpoint there :
image
You can see (L.92) that the `resolvePlaceholders` method won't be called, explaining my error since the raw string is passed on. 

Any idea on how we could fix this ? :)

lyca:
- should work in all Spring Boot applications
- i am unsure about potential class loading issues around the static EnvironmentHolder in the StaticEnvironmentProvider (it is working in my deployments without issues so far)
- the solution provided in #1375 for #361 seems to be overengineered
- fixes #1406


gregturn:
I'm not sure I see the benefit of moving to this bean-based approach of getting the `Environment` compared to simply pulling it out of the `WebApplicationContext` if needed.

lyca:
I currently do not see any method, that simply returns the `WebApplicationContext` for this to work. The current implementation does not work in Spring Boot as pointed out on SO over here https://stackoverflow.com/questions/22167912/contextloader-getcurrentwebapplicationcontext-always-returns-null and in issue #1406 by @adriendengreville

ShishkinDmitriy:
Hi guys, when it will be merged? 
Faced exactly the same problem in spring boot.

romanpierson:
Hi, is this ever to be merged? From reading https://github.com/spring-projects/spring-hateoas/pull/1328 I understood it was solved 2 years ago by @lyca and @odrotbohm stated its merged but for me its not working and can also not see those code changes in the latest source (using 2.6.9). Thanks for the clarification!

odrotbohm:
They don't need to be in there, as the fix for #1328 is actually a better solution than what's been suggested here. Would you mind creating a ticket and leave a few details of what exactly doesn't work, ideally with an accompanying reproducer.

romanpierson:
Sure I created a small reproducer - thanks for reaching out and let me know if that's not the intended way of how this should be used and work.

https://github.com/romanpierson/spring-hateoas-reproducer

ckdrunix:
> They don't need to be in there, as the fix for #1328 is actually a better solution than what's been suggested here. Would you mind creating a ticket and leave a few details of what exactly doesn't work, ideally with an accompanying reproducer.

As lyca pointed out #1328 did not resolve the issue for spring boot apps, since ContextLoader.getCurrentWebApplicationContext() is null in this case. Therefore PropertyResolvingMappingDiscovery.resolveProperties does not resolve the properties contained in the mapping (verified on spring boot 3.1.5, spring-hateoas 2.1.2). Therefore I think this PR is justified.

limminjeong98:
@odrotbohm  Hi, I want to know when this PR will be apply and release.

bcnapelinckx:
> > They don't need to be in there, as the fix for #1328 is actually a better solution than what's been suggested here. Would you mind creating a ticket and leave a few details of what exactly doesn't work, ideally with an accompanying reproducer.
> 
> As lyca pointed out #1328 did not resolve the issue for spring boot apps, since ContextLoader.getCurrentWebApplicationContext() is null in this case. Therefore PropertyResolvingMappingDiscovery.resolveProperties does not resolve the properties contained in the mapping (verified on spring boot 3.1.5, spring-hateoas 2.1.2). Therefore I think this PR is justified.

Completely correct... When will this fix be merged?

romanpierson:
Not sure what's the point of requesting a reproducer to be honest .......

PR #1475, opened 2021-02-19
Fix custom media type parser documentation

kylehoehns:
Noticed a slight error with the documentation while implementing my own custom media type for a project recently. Thanks!

### DIFF
--- a/src/main/asciidoc/mediatypes.adoc
+++ b/src/main/asciidoc/mediatypes.adoc
@@ -832,7 +832,7 @@ public class MyMediaTypeConfiguration implements HypermediaMappingInformation {
 
   @Override
   public List<MediaType> getMediaTypes() {
-    return MediaType.parse("application/vnd-acme-media-type") <1>
+    return MediaType.parseMediaTypes("application/vnd-acme-media-type"); <1>
   }
 
   @Override


gregturn (changes requested):
Actually, if it's a single media type, then `parseMediaType()` is the one you want. `parseMediaTypes` is for handling a comma-separated list of media types.

kylehoehns:
Thanks for the feedback @gregturn. The `HypermediaMappingInformation` interface declares a `List<MediaType> getMediaTypes()` method that consumers have to implement. It looks like `parseMediaType()` only returns a single `MediaType` so we'd need to put it into a collection before returning.

Would you rather I change it to something like this?

```java
return Collections.singletonList(MediaType.parseMediaType("application/vnd-acme-media-type"));
```

gregturn:
Thanks @kylehoehns for bringing this up. We'll take it from here.

gregturn:
Resolved via ce36feeb.

ingogriebsch:
@kylehoehns Would you mind to share with us which media-type you implement and if it will be available for the public? :)

kylehoehns:
@ingogriebsch I was implementing a custom media type for an internal application that just dealt with versioning. We ended up registering media types for `v1` and `v2` of our API and as a part of that I believe I needed to register those different version's media types so everything serialized/deserialized correctly. No fun public media types coming from me 😄

Issue #1563, opened 2021-06-29
Consider @JsonProperty-defined property name for affordance metadata


talentedasian:
Make the default name or value of the field `name` of the `DefaultMetadataProperty` to the value that is present in the **@JsonProperty** annotation else, use the default field name.

odrotbohm:
Would you mind taking a step back and describing the actual problem to be solved?

talentedasian:
![image](https://user-images.githubusercontent.com/71717525/123751314-3058b980-d8ea-11eb-8070-866a489be53b.png)
The image above is the desired output. Even though the field is named `politicalParty`, a `@JsonProperty` annotation is present with the value of `political_party`. This should be the default behavior for names of Hal+Form _templates property name value(`_templates.default.properties[0].name`) because the `@JsonProperty` values are the one that are expected in the `@RequestBody`. 

For example, I have a DTO that is used in a `@RequestBody`:

```java
public class AddPoliticianDTORequest {

        @JsonProperty("first_name")
	String firstName;

        @JsonProperty("last_name")	
	String lastName;
	
```
Spring MVC expects the body of the **POST** method to have json values of **first_name** and **last_name**. If I were to make an affordance for this in a Hal+Form mediatype, the resulting data would  be:
```
"templates":{
   "default":{
      "method":"post",
      "properties":[
         {
            "name":"politicalParty",
            "required":true,
            "type":"text"
         }
      ],
      "target":"http://localhost:8080/api/politician/politicians"
   }
}
The response would ultimately break your clients that use your api since you expect political_party instead of "politicalParty" all because of the @JsonProperty. The name of the property values should always be the value of a @JsonProperty annotation when it is present and if not, use the default field name.
```

odrotbohm:
Awesome, thanks for the additional context.

PR #1564, opened 2021-06-29
Add support for @JsonProperty annotation for DefaultMetadataProperty field "name".

talentedasian:
Make the default name of the DefaultMetadataProperty field "name" as the value of a @JsonProperty annotation given that the value is present, else, use the default implementation. 
Tests are located in the HalFormsTemplateBuilderUnitTest.

### DIFF
--- a/src/main/java/org/springframework/hateoas/mediatype/PropertyUtils.java
+++ b/src/main/java/org/springframework/hateoas/mediatype/PropertyUtils.java
@@ -448,7 +448,7 @@ public class PropertyUtils {
 		 */
 		@Override
 		public String getName() {
-			return property.getName();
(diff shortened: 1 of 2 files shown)


talentedasian:
Fixes #1563

talentedasian:
Closes #1563

odrotbohm:
@talentedasian – I have to ask you again. Please stop opening and closing tickets at will. We're organizing our work around this and having to follow up on all the notifications we get from ticket actions that appear for no obvious reason is time consuming. Please, just don't. Thanks!

odrotbohm:
I ended up choosing a slightly different approach to fixing that bug. Thanks for the contribution nonetheless.

PR #1567, opened 2021-06-29
Add support for @DecimalMin and @DecimalMax.

gregturn:


odrotbohm:
Wondering if there was a way to avoid the breaking change in the metadata APIs (`Long` -> `Number`). I've checked both the JSON:API and the Siren community media type implementations, and they don't even seem to call these methods. So, it might not be a very problematic issue.

gregturn:
I thought widening the type wasn’t considered breaking.

gregturn:
@odrotbohm Requested changes applied.

src/main/java/org/springframework/hateoas/AffordanceModel.java
```diff
@@ -149,8 +149,8 @@ public PayloadMetadata getOutput() {
 	 */
 	public <T> List<T> createProperties(BiFunction<InputPayloadMetadata, PropertyMetadata, T> creator) {
 
-		return input.stream()
-				.map(it -> creator.apply(input, it))
+		return input.stream() //
+				.map(it -> creator.apply(input, it)) //
```
cuspymd, line 153: I'm curious, what's the use of "//"?

odrotbohm: In a fluent API, in which you chain method calls, we use those to prevent the IDE from reformatting the code so that we end up with each invocation on a separate line.

odrotbohm:
> I thought widening the type wasn’t considered breaking.

Well, if you previously called that method, got a `Long` and now get a `Number`, that won't compile. Let's roll this into 1.4 M1 and see how many complaints we get. Not a candidate for back-porting then though.

src/main/java/org/springframework/hateoas/AffordanceModel.java
gregturn: Being managed by the Spring Data team, we format code on the backend via https://github.com/spring-projects/spring-data-build/tree/main/etc/ide.

We don't lean on maven plugins to reformat code nor ask that contributors "polish code" to fashion a pull request. Thus, this is how we maintain things.

odrotbohm:
Merged.

src/main/java/org/springframework/hateoas/AffordanceModel.java
cuspymd: Thanks for your detailed explanation.

Issue #1566, opened 2021-06-29
Consider @JsonProperty-defined property name for affordance metadata


odrotbohm:
Backport of #1563.

PR #1573, opened 2021-07-09
Consider supporting property names (see #1566) with @JsonNaming annotations.

talentedasian:
Add support for converting property names to cases e.g. camelCase with jackson's
@JsonNaming annotation.

### DIFF
--- a/src/test/java/org/springframework/hateoas/mediatype/PropertyUtilsTest.java
+++ b/src/test/java/org/springframework/hateoas/mediatype/PropertyUtilsTest.java
@@ -212,7 +212,7 @@ class PropertyUtilsTest {
 
 		InputPayloadMetadata metadata = PropertyUtils.getExposedProperties(JacksonPropertyNamingStrategyCustomizations.class);
 		
-		assertThat(getProperty(metadata, "SmallProperty")).isPresent();
+		assertThat(getProperty(metadata, "small_property")).isPresent();
 		assertThat(getProperty(metadata, "No_Name")).isPresent();
 	}
 	
@@ -222,7 +222,7 @@ class PropertyUtilsTest {
 		InputPayloadMetadata metadata = PropertyUtils.getExposedProperties(JacksonNullPropertyNamingStrategyCustomizations.class);
 		
 		assertThat(getProperty(metadata, "smallProperty")).isPresent();
-		assertThat(getProperty(metadata, "Name")).isPresent();
+		assertThat(getProperty(metadata, "name")).isPresent();
 	}
 
 	@Data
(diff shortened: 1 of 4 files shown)


talentedasian:
The main goal of the pull request is similar to of #1566,  To avoid discrepancies between Spring Hateoas' `Property Metadata` and Spring MVC's expectations on DTO names. With the `@JsonNaming` annotation in place, Spring MVC again expects the name of fields to be equal  to what Jackson will parse.

For example:

```java
@JsonNaming(PropertyNamingStrategies.SnakeCaseStrategy.class)
class DTO {
    String first_name;
    ......
}
```

```
{
  first_name: ....
}
```
Is expected and not "firstName" because of the `@JsonNaming` annotation in place.


That isn't the case with fields annotated with `@JsonProperty` though so `@JsonProperty` takes precedence over the `@JsonNaming` annotation.

src/test/java/org/springframework/hateoas/mediatype/PropertyUtilsTest.java
```diff
@@ -296,6 +320,20 @@ public void setFirstname(String firstname) {
...
+	@Value
+	@JsonNaming(PropertyNamingStrategies.UpperCamelCaseStrategy.class)
+	static class JacksonPropertyNamingStrategyCustomizations {
+		@JsonProperty("small_property") String smallProperty;
+		String no_Name;
+	}
+	
+	@Value
+	@JsonNaming
+	static class JacksonNullPropertyNamingStrategyCustomizations {
```
talentedasian, line 333: I would consider renaming this to `JacksonDefaultPropertyNamingStrategyCustomizations` for clarity.

odrotbohm:
Thanks for the contribution, but please refrain from "adding" changes on top of already fixed tickets. Also, it's always helpful to open a ticket for an aspect you'd like to see improved for discussion first, as that makes sure you don't spend time on some implementation that's unlikely to get accepted due to reasons that might come up in the discussion.

That said, I've considered this change and decided not to pursue that route. Jackson's (de)serialization customizations are applied to runtime instances of `ObjectMapper`. How those instances are configured and how that configuration affects the (de)serialization is guarded by the `OM`'s implementation details. That's why I don't think we should continue to bake more (de)serialization related aspects into the *statically computed* property metadata that is primarily driven by the logical view on the type to be serialized, the constraints imposed on the properties etc.

I am not certain that we can still do this in the 1.4 timeframe, but I would rather like to move to an abstraction that exposes (de)serialization related metadata based on the `OM` instance in use to the media-type-specific infrastructure, so that e.g. the code rendering HAL FORMS can access that metadata as bean, which in turn consolidates the property metadata with the Jackson (de)serialization settings.

I've file #1574 to keep track of those efforts.

PR #1576, opened 2021-07-12
AnnotationMappingDiscoverer should read @RequestMapping#params()

### LINKED ISSUE #467: AnnotationMappingDiscoverer should read @RequestMapping#params()
`@RequestMapping` allows to use requests' GET parameters to map incoming requests without having to inject parameters into a `@RequestParam` annotated method parameter:

```
@RequestMapping(method = GET, value = "/foobar", params = { "!one", "two", "three!=3", "four=4" })
public ResponseEntity<?> get() {
  return foobar();
}
```

This is pretty useful when you want clean API endpoints as well as clean code (e.g. no need of single method delegating to other methods thanks to a switch/case on the parameter's value).

It could be great if `linkTo(methodOn(FooBarApi.class).get())` took these `params` into account as well, by generating for the above example at least:

```
http://example.com/foobar?four=4
```

And why not:

```
http://example.com/foobar?four=4&two=true
```

But iff no `@RequestParam` matched this `two` parameter (otherwise, the actual value would have been given when calling `linkTo(methodOn(FooBarApi.class).get(twoParamValue))`).

What do you think about it?

Technically, when dealing with the `@RequestMapping` annotation, the `AnnotationMappingDiscoverer` instance used in `ControllerLinkBuilderFactory` could instantiate an new `org.springframework.web.servlet.mvc.condition.ParamsRequestCondition` to parse thoses `params`.

reda-alaoui:
Fixes #467

### DIFF
--- a/src/main/java/org/springframework/hateoas/server/core/AnnotationMappingDiscoverer.java
+++ b/src/main/java/org/springframework/hateoas/server/core/AnnotationMappingDiscoverer.java
@@ -15,6 +15,7 @@
  */
 package org.springframework.hateoas.server.core;
 
+import static java.util.Optional.ofNullable;
 import static org.springframework.core.annotation.AnnotatedElementUtils.*;
 import static org.springframework.core.annotation.AnnotationUtils.*;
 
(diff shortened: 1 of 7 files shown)


odrotbohm:
Thanks for the contribution. This is polished and merged!

PR #1754, opened 2022-01-18
Adds support for javax.validation.constraints.NotBlank

### LINKED ISSUE #1753: Add support for `javax.validation.constraints.NotBlank`
Currently, Jsr303AwarePropertyMetadata indicates that a property is required if the property is annotated with javax.validation.constraints.NotNull but it does not give the same consideration to properties annotated with javax.validation.constraints.NotBlank.

The presence of the `@NotBlank` annotation should be considered equivalent to the property being annotated with `@NotNull @Pattern(regex = "^\\s*(\\S+\\s*)+$")`.

Ideally, Jsr303AwarePropertyMetadata would support this annotation by updating its `isRequired` and `getPattern` methods. However, Jsr303AwarePropertyMetadata currently assumes that a property has zero-or-one `@Pattern` annotations, which is not valid. A property may have zero-or-more patterns.

In the case where a property is annotated with both `@NotBlank` and `@Pattern`, it may be acceptable to ignore the pattern implied by the `@NotBlank` annotation. Otherwise, it may be necessary to introduce a `getPatterns` method and deprecate `getPattern`.

fdutton:
closes #1753

### DIFF
--- a/src/main/java/org/springframework/hateoas/mediatype/PropertyUtils.java
+++ b/src/main/java/org/springframework/hateoas/mediatype/PropertyUtils.java
@@ -31,6 +31,7 @@ import javax.validation.constraints.DecimalMin;
 import javax.validation.constraints.Email;
 import javax.validation.constraints.Max;
 import javax.validation.constraints.Min;
+import javax.validation.constraints.NotBlank;
 import javax.validation.constraints.NotNull;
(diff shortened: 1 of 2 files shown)


odrotbohm:
That's polished and merged, thanks.

PR #1775, opened 2022-04-20
Optimization around constructing `UriTemplate` with multiple template variables of the same type

MikeRocke:
when supplying TemplateVariables of same Type.

This is done to reduce the repeated merge routines on ExpandGroups

I believe the current implementation degrades when more template variables are added, to demonstrate this
```
    @Benchmark
    @Fork(warmups = 1, value = 1)
    public void concat() {
        List<TemplateVariable> templateVariables = new LinkedList<>();
        for (int i = 0; i < 30; i++) {
            templateVariables.add(new TemplateVariable("a" + i, TemplateVariable.VariableType.REQUEST_PARAM));
        }
        TemplateVariables variables = new TemplateVariables(templateVariables);

        String uri = "http://localhost/autotrader/makes";
        UriTemplate template = UriTemplate.of(uri, variables);

    }
 ```
### Results

Before
```
Benchmark                Mode  Cnt     Score    Error  Units
BenchmarkRunner.concat  thrpt    5  8177.940 ± 55.365  ops/s
```

After

```

Benchmark                Mode  Cnt       Score       Error  Units
BenchmarkRunner.concat  thrpt    5  109527.885 ± 20025.995  ops/s
```

Hope this is ok and helpful! :)

### DIFF
--- a/src/main/java/org/springframework/hateoas/TemplateVariable.java
+++ b/src/main/java/org/springframework/hateoas/TemplateVariable.java
@@ -19,11 +19,7 @@ import static org.springframework.hateoas.TemplateVariable.VariableType.*;
 
 import java.io.Serializable;
 import java.nio.charset.StandardCharsets;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.List;
-import java.util.Map;
-import java.util.Objects;
+import java.util.*;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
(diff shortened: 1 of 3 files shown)


MikeRocke:
Closing this off for now. Happy to re-open if revisted

odrotbohm:
Sorry, this stayed unhandled for so long. Significant improvements. I've also additionally removed the streaming to group the variables in the first place and could bump the numbers by an additional 16%.

PR #1856, opened 2022-10-04
Add SlicedModel class for representing Slice collections

mschout:
This is groundwork needed to support SlicedResourcesAssembler<T> in spring-data-commons, by adding a SlicedModel class to spring hateoas.   I have another PR coming which adds the SlicedResourcesAssembler support to spring-data-commons.  Thi sis my first Spring contribution, so if I'm on the wrong track here, let me know.

See spring-projects/spring-data-commons#1307


odrotbohm:
That's polished and merged. Thanks for your contribution!

PR #2039, opened 2023-11-07
Support for HAL-FORMS value element

### LINKED ISSUE #1717: HalForm property value not set
For the example, I've got products with product references.

Here's the code to set "select-product" relation.

```
private static ProductDto withSelectProductAffordance(ProductDto productDto) {
        SelectProductRequest request = SelectProductRequest.builder()
                .reference(productDto.getReference())
                .build();

        Link selectProductLink = linkTo(methodOn(SelectProductResource.class)
                .selectProduct(request))
                .withRel(ProductCatalogRelations.SELECT_PRODUCT);

        Link affordance = selectProductLink
                .andAffordance(afford(methodOn(SelectProductResource.class)
                        .selectProduct(request)));

        return productDto.add(affordance);
    }
```

Here's the current result of the collection of products:

```
{
  "_embedded": {
    "products": [
      {
        "reference": "product-001",
        "_links": {
          "select-product": {
            "href": "http://localhost:8080/api/select-product"
          }
        },
        "_templates": {
          "default": { ... },
          "selectProduct": {
            "method": "post",
            "properties": [
              {
                "name": "reference",
                "required": true,
                "type": "text"
              }
            ],
            "target": "http://localhost:8080/api/select-product"
          }
        }
      },
      { ... }
    ]
  }
}
```

I would expect to have product reference already set in `selectProduct` template like below:

```
"properties": [
  {
    "name": "reference",
    "required": true,
    "type": "text",
    "value": "product-001"
  }
]
```

It seems that value does not exist in `AffordanceModel#PropertyMetadata` but is a valid in the spec: https://rwcbook.github.io/hal-forms/#_code_value_code

I would also expect the template name to be the same as the link rel.

reda-alaoui:
Fixes #1717

# Behaviour before this change

Let be:
```java

  class MyController {
    @GetMapping
    public ResponseEntity<?> list() {
      Link selfLink = selfLink.andAffordance(afford(methodOn(MyController.class).create(null)));
      return ResponseEntity.ok(new RepresentationModel<>(selfLink));
    }
  
    @PostMapping
    public ResponseEntity<?> create(@RequestBody Payload payload) {
      return ResponseEntity.created().build();
    }
  }
  
  record Payload(String foo, String bar) {
    Payload {
     // 'foo' and 'bar' are mandatory
     // we verify this invariant in the constructor to automatically end up with an HTTP 400 bad request in case of thrown exception
     requireNonNull(foo);
     requireNonNull(bar);
   }
  }
```

Calling GET will return something like:
```json
{
  "_templates": {
    "default": {
      "properties": [
        {
           "name": "foo"
         },
         {
           "name": "bar"
         }
      ]
    }
  }
}
```

We want to be able to assign a default value `hello` to attribute `foo` in the `HAL-FORMS` payload to obtain something like this:
```json
{
  "_templates": {
    "default": {
      "properties": [
        {
           "name": "foo",
           "value": "hello"
         },
         {
           "name": "bar"
         }
      ]
    }
  }
}
```

# Solution brought by this change

The implemented solution is heavily inspired from `HalFormsOptionsFactory`. A consumer can provide a value creator, taking a property metadata as input and returning a value of type `String` as output.

# Considered alternatives

## Retrieve the value directly from the payload instance

To do that, this kind of consumer code would be needed:
```java
@GetMapping
public ResponseEntity<?> list() {
    Link selfLink = selfLink.andAffordance(afford(methodOn(MyController.class).create(new Payload("hello", null))));
    return ResponseEntity.ok(new RepresentationModel<>(selfLink));
}
```

1. this will not work since `bar` is verified for non-nullity in `Payload` constructor. 
1. consumers will not either accept to relax the constructor invariant validation

On the 2nd point, some may argue that `jakarta.validation.constraints.NotNull` should be used instead of the in-house constructor validation. On that my opinion is as follow:
1. `@NotNull` should be avoided when plain simple java code is able to enforce the same constraint
1. IDEs will probably at least emit a warning when seeing a `null` value assigned to a  `@NotNull` property, at worst will fail the compilation

## Retrieve the value from a new annotation on the considered property

This would force the consumer to know the value before compilation.

### DIFF
--- a/src/main/java/org/springframework/hateoas/mediatype/hal/forms/HalFormsConfiguration.java
+++ b/src/main/java/org/springframework/hateoas/mediatype/hal/forms/HalFormsConfiguration.java
@@ -15,6 +15,7 @@
  */
 package org.springframework.hateoas.mediatype.hal.forms;
 
+import org.jspecify.annotations.Nullable;
 import tools.jackson.databind.json.JsonMapper.Builder;
 
 import java.util.ArrayList;
@@ -45,6 +46,7 @@ public class HalFormsConfiguration {
 	private final Map<Class<?>, String> patterns;
 	private final UnaryOperator<Builder> mapperCustomizer;
 	private final HalFormsOptionsFactory options;
+	private final HalFormsValueFactory values;
 	private final List<MediaType> mediaTypes;
 	private final boolean defaultSingleTemplate;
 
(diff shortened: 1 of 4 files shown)


kalgon:
Regarding the first considered alternative (_Retrieve the value directly from the payload instance_), maybe using 
 `jakarta.validation.constraints.NotNull` with a validation group could be an acceptable trade-off that won't trigger the IDE?

```java
public class Payload {

    // constructor, getters, setters

    @NotNull(groups = Submitted.class)
    private String foo;
    
    @NotNull(groups = Submitted.class)
    private String bar;
}

public class MyController {
 
    @GetMapping
    public ResponseEntity<?> list() {
        Link selfLink = selfLink.andAffordance(afford(methodOn(MyController.class).create(new Payload("hello", null))));
        return ResponseEntity.ok(new RepresentationModel<>(selfLink));
    }

    @PostMapping
    public ResponseEntity<?> create(@RequestBody @Validated(Submitted.class) Payload payload) {
        return ResponseEntity.created().build();
    }
}
```

reda-alaoui:
@kalgon , when I look at this, I see an annotation hell ^^

reda-alaoui:
@odrotbohm, could we have some feedback on this? I'd need the same thing for `HalFormsOptions` to be able to select the pre-selected option from the payload.

PR #2128, opened 2024-03-20
Fix link parsing for non-trivial cases

### LINKED ISSUE #2099: Support advanced Link header expressions
We've decided to use your library to parse `Link` headers, hoping that you'll correctly implement the specification intricacies, because doing parsing correctly is tricky. However, it's implemented completely naively, working only in the basic cases. It uses regex for parsing, even though it can't be used for non-context-free grammars.

Here are a bunch of examples:
```java
// should have failed
Link.valueOf("foo bar <url>;rel=\"next\"");

// greedy capture until the last `>`
System.out.println(Link.valueOf("<url>;customparam=foo>;rel=\"next\"").getHref());

// multiple rels not supported, should have been parsed to a collection
System.out.println(Link.valueOf("<url>;rel=\"next last\"").getRel().value());

// incorrect unquoting of values - missing end quote must be ignored, but currently "nex" is returned (missing last "t")
System.out.println(Link.valueOf("<url>;rel=\"next").getRel().value());

// param value not unescaped
System.out.println(Link.valueOf("<url>;rel=\"next\";title=\"foo\\\"bar\"").getTitle());

// incorrect semicolon handling - a semicolon within a quoted value is not treated literally
System.out.println(Link.valueOf("<url>;rel=\"next\";title=\"foo;bar\"").getRel().value());
```

This is the ABNF for the `Link` header:
```abnf
  Link       = #link-value
  link-value = "<" URI-Reference ">" *( OWS ";" OWS link-param )
  link-param = token BWS [ "=" BWS ( token / quoted-string ) ]
```

And here for the content of the `rel` param:
```abnf
  relation-type *( 1*SP relation-type )
where:

  relation-type  = reg-rel-type / ext-rel-type
  reg-rel-type   = LOALPHA *( LOALPHA / DIGIT / "." / "-" )
  ext-rel-type   = URI ; Section 3 of [RFC3986]
```

See https://httpwg.org/specs/rfc8288.html

So there can be multiple values in the `rel` param, even URIs and quoted URIs.

Some of the errors can't be fixed without breaking b-w compatibility, e.g. the decoding of href, unescaping of params and multiple rel values. Also, I'm not sure why custom parameters aren't supported - I didn't read the whole spec, but I don't think it prohibits custom params (we don't use them, just wondering).

viliam-durina:
The previous implementation used naive parsing suffering from many issues, especially when special characters were part of values. It also didn't escape the special characters when serializing a link to string.

I marked the `Link.valueOf` method as deprecated because it doesn't correctly parse multiple links, nor does it handle multiple values for the `rel` param. One should use `Links.parse`. There are no other incompatible API changes.

I had to change one current test: we no longer ignore missing final `>` after the URL - such link is invalid anyway because it didn't have the rel parameter.

Fixes #2099

Tests have 100% coverage of the new code, except for one line that cannot currently happen, but is there for robustness.

### DIFF
--- a/src/main/java/org/springframework/hateoas/Link.java
+++ b/src/main/java/org/springframework/hateoas/Link.java
@@ -23,17 +23,16 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
 
+import org.springframework.lang.NonNull;
 import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
-import org.springframework.util.StringUtils;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import org.springframework.util.StringUtils;
 
 /**
  * Value object for links.
@@ -41,13 +40,13 @@ import com.fasterxml.jackson.annotation.JsonProperty;
  * @author Oliver Gierke
  * @author Greg Turnquist
  * @author Jens Schauder
+ * @author Viliam Durina
  */
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonIgnoreProperties(value = { "templated", "template" }, ignoreUnknown = true)
 public class Link implements Serializable {
 
 	private static final long serialVersionUID = -9037755944661782121L;
-	private static final Pattern URI_AND_ATTRIBUTES_PATTERN = Pattern.compile("<(.*)>;(.*)");
 
 	public static final String ATOM_NAMESPACE = "http://www.w3.org/2005/Atom";
 
@@ -241,7 +240,7 @@ public class Link implements Serializable {
 	}
 
 	/**
-	 * Creats a new {@link Link} with the given {@link Affordance}s.
+	 * Creates a new {@link Link} with the given {@link Affordance}s.
 	 *
 	 * @param affordances must not be {@literal null}.
 	 * @return will never be {@literal null}.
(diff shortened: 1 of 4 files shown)


odrotbohm:
Given this adds significant complexity, would you mind extracting the actual parsing into a package protected type `LinkParser`? Link could then use that functionality from `….valueOf(…)`. It might make sense to move all (previously existing) parsing related tests into a dedicated unit test class solely focusing on that. Furthermore, please make sure to apply the formatters provided [here](https://github.com/spring-projects/spring-hateoas/tree/main/etc/ide) and avoid unnecessary formatting changes (e.g. the removal of blank lines in the tests). Finally, please squash the individual commits into a single commit.

viliam-durina:
@odrotbohm I'll work on that. I have just one question: are the blank lines part of your standard format? I normally fix format of the code _near_ which I'm working.

viliam-durina:
I pushed the requested changes, except:

- I didn't move the parsing-related tests to a separate class. I test using the public API, which is the right thing to do IMO. And the public api is the `Links` class, not in the package-private `LinkParser` class. If you want to test against the internal API, or if you want to test the public API from a test class not strictly following the convention of appending `UnitTest` to the tested class name, let me know.

- I'm using Intelij, but I don't seem to be able to import the formating rules in a `jar`. I ended importing the Eclipse XML rules to IntelliJ

- I think the formatting after after auto-formatting the code is much less readable. Especially the line breaks in multi-line expressions are now placed randomly, probably after reaching a certain line length, and the result is much harder to read, especially the tests. Are you sure you want to apply the formatting? Could I get an exemption at least on the line breaks?

- I rebased the branch (was quite stale), but I didn't squash yet. IMO squashing makes it much harder for reviewers to follow the changes. Furthermore, I hope you'll reconsider the formatting requirement, and I want to be able to revert that commit, which wouldn't be possible if I squashed them. There's a squash option when you'll be merging the PR, that's much more useful. You can even set github up so that a squash merge is the only option, we do that in all our projects and generally frown on force-pushes.

odrotbohm:
I'll take it from here.

odrotbohm:
That's merged, thanks!

PR #2305, opened 2025-04-14
refactor: Use `Collection#isEmpty()` instead of comparing `size()`

greg-at-moderne:
**What**:
Changing calls to `Collections.size() == 0` to `Collection.isEmpty()`.

**Why**:
This is more readable and likely also more efficient.

**How**:
This has been generated using OpenRewrite's https://docs.openrewrite.org/recipes/staticanalysis/isemptycalloncollections


odrotbohm:
That's merged, thanks!

Issue #1240, opened 2020-03-27
Add support for @ModelAttribute


Gearme:
Controller method parameters annotated with MVC's `@ModelAttribute` currently do not get represented in HAL. Adding support for that Annotation (provided explictly or [implicitly](https://docs.spring.io/spring/docs/current/spring-framework-reference/web.html#mvc-ann-arguments)) would be appreciated. @gregturn suggested that `QueryParameter`, currently responsible for `@RequestParam`, could be modified to cover that Attribute as well.

Note that [RFC6570 ](https://tools.ietf.org/html/rfc6570) - aside from its [form-style query expansion](https://tools.ietf.org/html/rfc6570#section-3.2.8) - also defines a [composite value](https://tools.ietf.org/html/rfc6570#section-2.4.2) using an "explode" modifier ("*"). Support for `@ModelAttribute` could conveivably take either shape, or even offer that choice to the developer in some way.

|style|example|
| ---- | ---- |
|simple form style|`{?field1,field2,field3,page,size,sort}`|
|with composite value|`{?myClass*,page,size,sort}`|

gabbloquet:
Hi team,
Thanks for this amazing work !
Any news on this subject ? :) 
Have a nice day !

PR #2360, opened 2025-09-08
GH-1240 - Add support for @ModelAttribute.

snowykte0426:
Adds support for Spring MVC's `@ModelAttribute` in link building, resolving #1240.

Spring MVC binds a `@ModelAttribute` parameter from the individual request parameters matching the properties of its type. Link building now mirrors that, using RFC6570's form-style query expansion, with unpopulated properties left as template variables:

```
linkTo(methodOn(C.class).search(null))                ->  /search{?category,sortBy,tags}
linkTo(methodOn(C.class).search(partiallyPopulated))  ->  /search?category=books{&sortBy,tags}
linkTo(methodOn(C.class).search(fullyPopulated))      ->  /search?category=books&sortBy=name&tags=a&tags=b
```

Implemented through the existing extension points — `@ModelAttribute` is registered in `WebHandler`'s `HandlerMethodParameters.ANNOTATIONS` and `HandlerMethodParameter.FACTORY` alongside `@RequestParam` and `@PathVariable`. Since `WebHandler` backs both stacks, WebFlux is covered too. No new public API.

Only parameters explicitly annotated with `@ModelAttribute` are considered, and only the form-style expansion is implemented. See the discussion below for the reasoning.

### DIFF
--- a/src/main/asciidoc/server.adoc
+++ b/src/main/asciidoc/server.adoc
@@ -138,6 +138,61 @@ assertThat(link.getHref()).endsWith("/people?names=Matthews,Beauford"); <3>
 NOTE: The reason we're exposing `@NonComposite` is that the composite way of rendering request parameters is baked into the internals of Spring's `UriComponents` builder and we only introduced that non-composite style in Spring HATEOAS 1.4.
 If we started from scratch today, we'd probably default to that style and rather let users opt into the composite style explicitly rather than the other way around.
 
+[[server.link-builder.webmvc.methods.model-attributes]]
+==== Building links for `@ModelAttribute` parameters
+
+Spring MVC binds a handler method parameter annotated with `@ModelAttribute` from the individual request parameters that match the properties of the attribute's type.
+Link building mirrors that: each bindable property contributes its own request parameter, following https://tools.ietf.org/html/rfc6570#section-3.2.8[RFC6570's form-style query expansion].
+Properties you have not populated are rendered as template variables, so that a client can fill them in.
+
+====
+[source, java]
+----
+class Filter { <1>
+
+  String category;
+  String sortBy;
+
+  // getters and setters
+}
+
+@Controller
+class PersonController {
+
+  @GetMapping("/people")
+  HttpEntity<PersonModel> showAll(@ModelAttribute Filter filter) { … }
+}
+
+var link = linkTo(methodOn(PersonController.class).showAll(null)).withSelfRel(); <2>
+
+assertThat(link.getHref()).endsWith("/people{?category,sortBy}");
+
+var filter = new Filter();
+filter.setCategory("customer");
+
+link = linkTo(methodOn(PersonController.class).showAll(filter)).withSelfRel(); <3>
+
+assertThat(link.getHref()).endsWith("/people?category=customer{&sortBy}");
+----
+<1> A plain form backing object. Java records work just as well, as their components are bound through the canonical constructor.
+<2> Invoking the method without an attribute renders all bindable properties as template variables.
+<3> Populated properties are rendered as actual request parameters, the remaining ones stay templated.
+====
+
+A property is considered bindable if it is writable (or a record component) and resolves to a simple type or a collection of such.
+`Collection`-valued properties are rendered in the composite style, i.e. the parameter name is repeated for each value; arrays are rendered comma-separated, matching how `@RequestParam` arrays already render.
+Derived, read-only properties and nested objects are skipped, as is the entire attribute if it is declared as `@ModelAttribute(binding = false)`.
+A property whose name is already taken by a `@RequestParam` or `@PathVariable` of the same method is skipped too, so that the declared parameter wins and the link stays expandable.
+
+NOTE: A property of a primitive type can never be rendered as a template variable once the attribute itself is present, because a primitive cannot express "unset" — `int page` is indistinguishable from `page = 0`.
+Such properties are always rendered as bound request parameters.
+Use the boxed type (`Integer`) if you want the property to stay fillable by the client.
+
+NOTE: Only parameters explicitly annotated with `@ModelAttribute` are considered.
+Spring MVC additionally treats unannotated complex parameters as implicit model attributes, but it can only do so because its model attribute resolver runs after every other argument resolver has had a chance to claim the parameter.
+Link building has no such chain to defer to, which means an unannotated parameter cannot be told apart from the likes of `HttpServletRequest`, `Pageable` or `BindingResult`.
+Annotate the parameter explicitly to have it rendered into the link.
+
 [[server.link-builder.webflux]]
 == Building links in Spring WebFlux
 
--- a/src/main/java/org/springframework/hateoas/QueryParameter.java
+++ b/src/main/java/org/springframework/hateoas/QueryParameter.java
@@ -25,6 +25,7 @@ import org.springframework.core.annotation.MergedAnnotations;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.ModelAttribute;
 
 /**
  * Representation of a web request's query parameter (https://example.com?name=foo) => {"name", "foo", true}.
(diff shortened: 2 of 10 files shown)


snowykte0426:
I've rebased this onto current `main` and reworked the implementation from scratch. The previous version had a few problems I want to be upfront about:

- The link URI template was never actually affected. Template generation happens in `WebHandler`, which the old change didn't touch, so `linkTo(methodOn(…))` produced the exact same href as before.
- The `exploded` flag added to `QueryParameter` was not read by any production code, so the `{?myClass*}` form advertised in the PR description was never rendered anywhere.
- Sweeping all method parameters and treating unannotated complex types as implicit `@ModelAttribute`s pulled `HttpServletRequest`, `BindingResult`, `Pageable` and friends into the affordance query parameters.
- The `SpringAffordanceBuilderModelAttributeTest` never passed — `getAffordanceModel(APPLICATION_JSON)` returns `null`, so all four tests errored with an NPE.

Apologies for the noise on the first attempt. Here's what the branch does now.

### Approach

Rather than introducing a parallel mechanism, this uses the extension points that were already there:

- **`WebHandler`** — `@ModelAttribute` is registered in `HandlerMethodParameters.ANNOTATIONS` and `HandlerMethodParameter.FACTORY`, alongside `@RequestParam` and `@PathVariable`, with a `ModelAttributeParameter` implementation. This is what makes the URI template actually change.
- **`ModelAttributeProperties`** (new, package private) — detects the bindable properties of the attribute type, cached.
- **`SpringAffordanceBuilder`** — expands a `@ModelAttribute` parameter into one `QueryParameter` per property. The existing `@RequestParam` handling is untouched.
- **`QueryParameter`** is not modified at all, so no new public API.

### Behaviour

Spring MVC binds a `@ModelAttribute` from the individual request parameters matching the properties of its type, and link building now mirrors that — RFC6570 form-style query expansion, with unpopulated properties left as template variables:

```
linkTo(methodOn(C.class).search(null))                  ->  /search{?category,sortBy,tags}
linkTo(methodOn(C.class).search(partiallyPopulated))    ->  /search?category=books{&sortBy,tags}
linkTo(methodOn(C.class).search(fullyPopulated))        ->  /search?category=books&sortBy=name&tags=a&tags=b
linkTo(methodOn(C.class).mixed("q1", null))             ->  /mixed?q=q1{&category,sortBy,tags}
```

`@ModelAttribute(binding = false)` contributes nothing. The same properties are exposed as query parameters on the affordances built for the method.

A property counts as bindable if it is writable (or a record component, since those bind through the canonical constructor) and of a simple type or a collection of such. Collection-valued properties render in the composite style, consistent with `@RequestParam`. Derived read-only properties and nested objects are skipped. Property order is sorted, because `Introspector` does not return a stable order for records.

### Two deliberate omissions

**Implicit `@ModelAttribute`.** The issue mentions it, but I left it out on purpose. Spring MVC can resolve unannotated complex parameters as model attributes only because `ModelAttributeMethodProcessor` is registered after every other argument resolver has had a chance to claim the parameter. Link building has no such chain to defer to, so an unannotated parameter cannot be distinguished from `HttpServletRequest`, `Pageable`, `BindingResult` and the rest — which is exactly how the previous version broke. Supporting it seems to need an explicit opt-in of some kind, and that felt like your call to make rather than mine. Happy to add it if you'd like to point me at a shape you'd accept.

**The composite `{?myClass*}` form.** Only the form-style expansion from the issue's table is implemented. `TemplateVariable` already carries `Cardinality.COMPOSITE`, so that variant can be layered on later if wanted, but expanding a POJO as a composite value doesn't give a client anything it can fill in, whereas the per-property form matches what the binder actually consumes.

### Tests and docs

`WebMvcLinkBuilderUnitTest` covers the rendered link templates, `ModelAttributePropertiesUnitTest` the property detection rules, and `SpringAffordanceBuilderUnitTest` the affordance query parameters — including regression guards asserting that unannotated parameters and `binding = false` contribute nothing. Full suite is green at 997 tests. `server.adoc` has a new section under "Building links that point to methods", including a note explaining the implicit-attribute limitation.

Feedback very welcome, particularly on the two omissions above.

snowykte0426:
I put this branch through another round of review and found enough real problems that I've pushed a reworked version. Summary of what was wrong and what changed.

### Dropped: the affordance half

The previous push also expanded `@ModelAttribute` into `QueryParameter`s in `SpringAffordanceBuilder`. I checked what that actually produced in serialized output, and it was not worth having:

- **HAL-FORMS** — inert. `HalFormsTemplateBuilder` filters GET affordances out, and for non-GET it builds properties from `getInput()`, never from `getQueryMethodParameters()`. The data was in the model but nothing rendered it.
- **Collection+JSON** — dead. `CollectionJsonJacksonModule.findQueries` only admits non-GET affordances, while `CollectionJsonAffordanceModel.determineQueryProperties()` only returns anything for GET. The two conditions are mutually exclusive, so `queries[].data` is always empty. (Pre-existing, but it made the addition dead code.)
- **UBER** — visible, but it caused a regression: `UberData.mergeDeclaredLinksIntoAffordanceLinks` inner-joins declared and affordance links on the URL string, and templating the declared href broke the match, so the link disappeared from the output entirely.

So it was inert in two of three consumers and net-negative in the third. `SpringAffordanceBuilder` is untouched again and the diff is smaller for it.

### Fixed

1. **NullAway violation** — `bindModelAttributeParameters` dereferenced a `@Nullable` `BeanWrapper`. The `nullaway` profile failed to compile. Restructured with an early return; that profile is now back to the single pre-existing error on `main` (`LinkBuilderSupport:114`).
2. **Name collisions produced contradictory URIs.** A `@ModelAttribute` property sharing a name with a `@RequestParam`, a `@PathVariable`, or a property of a second `@ModelAttribute` was emitted twice — and in the worst case a name that was already bound was *also* advertised as a template variable (`?category=fromForm{&category}`), which cannot round-trip. Names are now reserved as they are bound and a colliding property is skipped, so the explicitly declared parameter wins.
3. **A throwing getter broke link building.** `BeanWrapper.getPropertyValue` was called unguarded, so any form object with a lazily-computing or defensive writable getter turned link assembly into an exception — a failure mode that did not exist before this change. A failing read is now treated as an absent value.
4. **Type-variable properties were silently dropped.** `isBindable` looked at the erased `PropertyDescriptor.getPropertyType()`, so `T value` on `Form<T>` never rendered even when populated. Property types are now resolved against the owning `ResolvableType`, and the cache is keyed by `ResolvableType` so `Form<String>` and `Form<LocalDate>` are told apart.

### Documented rather than fixed

**Primitive properties.** A primitive cannot express "unset", so once the attribute itself is non-null, `int page` always renders as `page=0` and can never be a template variable. There is no way to distinguish the default from an intentional zero. This is now stated in the reference docs with the suggestion to use the boxed type, and pinned with a test.

**Array properties** render comma-joined rather than exploded. That matches existing `@RequestParam String[]` behaviour and the binder parses it back, so I kept the behaviour and corrected the docs, which previously claimed otherwise.

### Tests

Added coverage for the four fixes above plus records at the rendered-href level, and — since `WebHandler` backs both stacks — `@ModelAttribute` cases in `WebFluxLinkBuilderTest`, which had none. Full suite is green at 1004.

The two deliberate omissions are unchanged: implicit `@ModelAttribute`, and the composite `{?myClass*}` form. Reasoning for both is in the earlier comment.

snowykte0426:
This has been rebased onto current `main` and the implementation reworked from scratch. The original version of this PR was wrong in ways worth stating plainly: it only touched `QueryParameter`, so the link URI template never actually changed; the `exploded` flag it added had no production readers, so the `{?myClass*}` form it advertised was never rendered; it swept every method parameter and treated unannotated complex types as implicit model attributes, which pulled `HttpServletRequest`, `BindingResult` and friends into the affordance query parameters; and its own `SpringAffordanceBuilderModelAttributeTest` never passed. Apologies for that — none of it is on the branch any more.

### What it does now

Spring MVC binds a `@ModelAttribute` parameter from the individual request parameters matching the properties of its type. Link building mirrors that, using RFC6570 form-style query expansion, with unpopulated properties left as template variables:

```
linkTo(methodOn(C.class).search(null))                ->  /search{?category,sortBy,tags}
linkTo(methodOn(C.class).search(partiallyPopulated))  ->  /search?category=books{&sortBy,tags}
linkTo(methodOn(C.class).search(fullyPopulated))      ->  /search?category=books&sortBy=name&tags=a&tags=b
```

I verified the round trip end to end: expanding the rendered template via `Link.expand(...)` and replaying the result against a live MVC handler binds every property back into the command object, collections included.

### Approach

Rather than a parallel mechanism, this uses the extension points that were already there:

- **`WebHandler`** — `@ModelAttribute` is registered in `HandlerMethodParameters.ANNOTATIONS` and `HandlerMethodParameter.FACTORY` alongside `@RequestParam` and `@PathVariable`, with a `ModelAttributeParameter` implementation. This is what makes the URI template actually change. Since `WebHandler` backs both stacks, WebFlux is covered too.
- **`ModelAttributeProperties`** (new, package private) — detects bindable properties, cached by `ResolvableType`.
- **`QueryParameter` and `SpringAffordanceBuilder` are untouched**, so there is no new public API.

On the affordance side: I did have `SpringAffordanceBuilder` expanding `@ModelAttribute` into `QueryParameter`s, and dropped it after checking what it produced in serialized output. It is inert in HAL-FORMS (`HalFormsTemplateBuilder` skips GET affordances and builds properties from `getInput()`, never from `getQueryMethodParameters()`); dead in Collection+JSON (`findQueries` admits only non-GET while `determineQueryProperties()` returns values only for GET, so `queries[].data` is always empty); and in UBER it caused a regression, because `UberData.mergeDeclaredLinksIntoAffordanceLinks` inner-joins declared and affordance links on the URL string and templating the declared href broke the match, dropping the link from the output. Inert in two of three consumers and net-negative in the third, so it is gone.

### Behaviour worth knowing

A property is bindable if it is writable (or a record component, as those bind through the canonical constructor) and resolves to a simple type or a collection of such. Type variables are resolved against the owning type, so `T value` on `Form<T>` renders according to the type argument actually used. Derived read-only properties and nested objects are skipped, as is the whole attribute under `@ModelAttribute(binding = false)`.

A property whose name is already taken by a `@RequestParam` or `@PathVariable` on the same method is skipped, so the declared parameter wins. Without that, a bound name could also be advertised as a template variable (`?category=fromForm{&category}`), which cannot round-trip. Worth a second opinion if you would rather the attribute won instead.

A failing getter is treated as an absent value rather than propagating, so a form object with a lazily-computing or defensive getter cannot turn link assembly into an exception.

Two limitations are documented rather than fixed. A primitive property cannot express "unset", so `int page` always renders as `page=0` once the attribute is present; the reference docs say so and suggest the boxed type. Array properties render comma-joined rather than exploded, which matches existing `@RequestParam String[]` behaviour and parses back correctly.

### Deliberately out of scope

**Implicit `@ModelAttribute`.** The issue asks for it, but Spring MVC can resolve unannotated complex parameters only because `ModelAttributeMethodProcessor` is registered last, after every other argument resolver has had its chance. Link building has no such chain to defer to, so an unannotated parameter cannot be told apart from `HttpServletRequest`, `Pageable`, `BindingResult` and anything a custom resolver claims — which is precisely how the first version of this PR broke. A denylist cannot fix that; `Pageable` in particular already has correct handling via a `UriComponentsContributor`, so a miss would double-render it. Requiring the explicit annotation seems like the only defensible option, and the reference docs explain why.

**The composite `{?myClass*}` form.** Only the form-style expansion is implemented. `TemplateVariable` already carries `Cardinality.COMPOSITE`, so it could be layered on later, but the composite form names a server-side type the client cannot enumerate, cannot express partial population (there is no equivalent of `?category=books{&sortBy}`), and would collapse the affordance to a single useless parameter. Happy to add an opt-in if you disagree.

### Tests and docs

`WebMvcLinkBuilderUnitTest` covers the rendered link templates including records, generics, name collisions, failing getters, arrays and the primitive limitation; `ModelAttributePropertiesUnitTest` covers the property detection rules; `WebFluxLinkBuilderTest` gains `@ModelAttribute` cases, having had none. Regression guards assert that unannotated parameters and `binding = false` contribute nothing. Full suite green at 1004, and the `nullaway` profile is back to the single pre-existing error on `main` (`LinkBuilderSupport:114`).

`server.adoc` has a new section under "Building links that point to methods" documenting the behaviour and both limitations.

@odrotbohm — this has been sitting a while and no longer resembles what was originally submitted, so it may be worth a fresh look rather than reading the old diff. Two things I would especially value your call on: whether the declared `@RequestParam`/`@PathVariable` should win a name collision the way it does here, and whether requiring the explicit annotation instead of also supporting implicit model attributes is an acceptable scope for a first cut. Happy to keep iterating.

PR #2430, opened 2026-02-13
Support HAL-Forms options 'selectedValues'

### LINKED ISSUE #2400: Missing documentation / bug? - What is HalFormsOptions.selectedValue about?
The specification mentions `property.value` (to express a single, selected value) and `property.options.selectedValues` to express a list of selected values.

The `HalFormsOptions.selectedValue` field seems to be neither of those things?

The field seems to become `property.value` (singular) in the template, but in that case the naming seems off (it should not be part of the "property.options", but it should be directly on the `property`. The `property.options` only has the `selectedValues` field for the values, according to the specification.

The javadoc for the `HalFormsOptions.getSelectedValue()` is also missing, making it hard to guess the intent of this field. Is this an oversight?

It would be great to have a `HalFormsOptions.getSelectedValues()` that behaves according to the specification and maybe the current`HalFormsOptions.getSelectedValue()` should be deprecated and replaced by something else, that is related to the property more clearly (instead of sitting on the options sub-object).

reda-alaoui:
Fix #2400

### DIFF
--- a/src/main/java/org/springframework/hateoas/mediatype/hal/forms/HalFormsOptions.java
+++ b/src/main/java/org/springframework/hateoas/mediatype/hal/forms/HalFormsOptions.java
@@ -63,7 +63,7 @@ public interface HalFormsOptions {
 
 		Assert.notNull(values, "Values must not be null!");
 
-		return new Inline(values, null, null, null, null, null);
+		return new Inline(values, null, null, null, null, null, null);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


reda-alaoui:
Looks like the CI build can never succeed:
```
Caused by: java.io.IOException: JFrog service failed. Received 401: {
  "errors" : [ {
    "status" : 401,
    "message" : "Artifactory configured to accept only encrypted passwords but received a clear text password, getting the encrypted password can be done via the WebUI."
  } ]
}
```

